PR #29, opened 2016-12-29
Keep scroll postion.

ivarconr:
/cc @sveisvei

### DIFF
--- a/src/component/app.jsx
+++ b/src/component/app.jsx
@@ -93,12 +93,19 @@ export default class App extends Component {
     }
 
     render () {
+        const shouldUpdateScroll = (prevRouterProps, { location }) => {
+            if (prevRouterProps && location.pathname !== prevRouterProps.location.pathname) {
+                return location.action === 'POP';
+            } else {
+                return [0, 0];
+            }
+        };
         const createListItem = (path, caption, icon) =>
-            <a
-                href={this.context.router.createHref(path)}
+            <Link
+                to={path}
                 className={this.context.router.isActive(path) ? style.active : ''}>
                 {icon && <Icon name={icon} />} {caption}
-            </a>;
+            </Link>;
 
         return (
             <div style={{}}>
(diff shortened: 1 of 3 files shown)


src/component/app.jsx
```diff
@@ -42,10 +43,6 @@ export default class App extends Component {
         if (this.props.location.pathname !== nextProps.location.pathname) {
             clearTimeout(this.timer);
             this.timer = setTimeout(() => {
-                window.requestAnimationFrame(() => {
-                    document.querySelector('.mdl-layout__content').scrollTop = 0;
```
ivarconr: @sveisvei Why did you add this in the first place?

src/component/app.jsx
```diff
@@ -122,6 +119,7 @@ export default class App extends Component {
                             {createListItem('/applications', 'Applications', 'apps')}
                         </Navigation>
                     </Drawer>
+                    <ScrollContainer scrollKey="main">
```
ivarconr: seems to work with same key for everything. Might want to use the current path (window.location.pathname)?

src/component/app.jsx
sveisvei: Because navigating a new page should send you to top.

src/component/app.jsx
sveisvei: makes sense to lock to the pathname.

sveisvei:
wfm (when using pathname instead of "main")

ivarconr:
Kind of works. But feels strange to keep scroll position when using menu and not "back button".. @SimenB ?

SimenB:
Can you check the navigation event? If it's go back (`POP`?) don't scroll?

sveisvei:
Agree, it does not work out great when not using back. In general we just want it in list views.

sveisvei:
btw, until we have tests we should just run the build - would have caught the missing dependency.

ivarconr:
i can add that

ivarconr:
scrollTo uses pathname under the hood to use correct scrolling per path. I don't see any reason to not have the same scrollKey in the ScrollContainer, it only makes sense if we have multiple scrollable containers on the same pathname.

ivarconr:
It seems to be working now with a custom `shouldUpdateScroll`. It only scroll down when it sees a "POP" action.

SimenB:
Nice! 🎉

sveisvei:
wfm

sveisvei:
💯

Issue #30, opened 2016-12-30
Add drag n drop reorder strategies


sveisvei:

ivarconr:
Yes ;)

sveisvei:
I will have a go at this, will use `react-dnd` if nothing better comes to mind.

PR #31, opened 2016-12-30
Sort strategies

sveisvei:
Closes https://github.com/Unleash/unleash-frontend/issues/30

I did not add a hover-effect to the drop-target (yet)


ivarconr:
Cool 💎⛓

PR #34, opened 2017-01-02
make link to feature toggle span full width of list

vsandvold:
Added 100% width to feature link + some minor refactoring making use of mdl css classes.

### DIFF
--- a/src/component/feature/feature-list-item-component.jsx
+++ b/src/component/feature/feature-list-item-component.jsx
@@ -32,7 +32,7 @@ const Feature = ({
 
     return (
         <li key={name} className="mdl-list__item">
-            <span className="mdl-list__item-primary-content">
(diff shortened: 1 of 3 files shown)


sveisvei:
lgtm

ivarconr:
Cool, and welcome to the party @vsandvold 🎉

sveisvei:
Very long names with many strategies(> 3) does create some issues here @vsandvold. Not sure how we can solve this in a nice way.

sveisvei:
https://github.com/Unleash/unleash-frontend/pull/35 might help with the chips at least

vsandvold:
Have you considered two lines per feature item (using `mdl-list__item--two-line`)? That would give more space for strategies. I can set up a PR with a suggestion if you like.

PR #46, opened 2017-01-08
summarize strategies for toggles with more than 3

vsandvold:
Cuts back on some of the content to make more space for everything else.

@sveisvei Do you think we can combine this with regular chips as well, with regular font size and spacing?


sveisvei:
👍

Issue #47, opened 2017-01-17
Move github link and username into menu


vsandvold:
I suggest we move the github link and username from the header into the menu drawer. It will clean up and make more space in the header for other, perhaps more important elements.

Suggestion:
![fireshot capture 205 - unleash admin - http___localhost_3000_ _features_view_asdfasdf](https://cloud.githubusercontent.com/assets/6734787/22040207/6076d586-dd02-11e6-88f0-ad65f71b540e.png)

sveisvei:
Sure thing, wfm - but the username might be in place also - imo. But not important now - we support richer profile-data (name + image) etc via Okta soon as well maybe?

vsandvold:
No prob, I can leave the username where it is now. A logo would be nice too, at the top of the menu drawer :-)

SimenB:
👍 Please call it "GitHub" as well

PR #51, opened 2017-01-19
github link into menu + username tooltip + visual tweaks

vsandvold:
#47 

Before:
fireshot capture 212 - unleash admin - http___localhost_3000_ _features

After:
fireshot capture 211 - unleash admin - http___localhost_3000_ _features


ivarconr:
some linting errors, besides from that this should be good to merge!

Issue #50, opened 2017-01-17
New feature toggle page layout


vsandvold:
Desktop:
![fireshot capture 199 - unleash admin - http___localhost_3000_ _features_view_asdfasdf](https://cloud.githubusercontent.com/assets/6734787/22041592/bcf5a9f4-dd07-11e6-8f17-12bac06516b9.png)

Mobile:
![fireshot capture 200 - unleash admin - http___localhost_3000_ _features_view_asdfasdf](https://cloud.githubusercontent.com/assets/6734787/22041604/c71d4a4a-dd07-11e6-992f-d40c9b198b90.png)

ivarconr:
This is definitely better than our current implementation. See no reason not to move in this direction.

sveisvei:
This improvement is a no-brainer indeed.

Issue #19, opened 2016-12-20
Improve visual presntation of "History view"


ivarconr:

ivarconr:
It also looks a bit old when you "enabled/disables" a toggle and it is not directly updated in the history tab of a feature toggle.

sveisvei:
In general we should just make the data-fetching via a socket with events, or else its just monkeypatching imo.

ivarconr:
Yes in a perfect world. Will require some extra work on the API, and I want
to do that, but not as part of this issue.

PR #55, opened 2017-01-25
added data table header component with title and actions

vsandvold:
Preparation for #50 , but also relevant for #19 ?

Implements a data table header from the material design spec: https://material.io/guidelines/components/data-tables.html#data-tables-tables-within-cards

A footer with pagination would also be nice :-)

Before:
fireshot capture 218 - event history unleash - http___localhost_3000_ _history

After:
fireshot capture 220 - event history unleash - http___localhost_3000_ _history

### DIFF
--- a/src/component/common/common.scss
+++ b/src/component/common/common.scss
@@ -8,6 +8,15 @@
     width: 100%;
 }
 
+.horizontalScroll {
+    overflow-x: scroll;
+    -webkit-overflow-scrolling: touch;
+}
+
+.horizontalScroll::-webkit-scrollbar {
+    display: none;
+}
+
 .divider {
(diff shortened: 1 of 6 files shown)


src/component/common/common.scss
```diff
@@ -22,3 +31,36 @@
         display: none;
     }
 }
+
+.dataTableHeader {
```
ivarconr: to bad this is not part of react-mdl.

ivarconr:
nice job! =)

src/component/common/common.scss
vsandvold: I know, and it's not part of MDL either.

PR #56, opened 2017-01-26
refactor layout into a new form add component

vsandvold:
Another small preparation for #50 , just moving specific layout for the create feature toggle page into a separate component.


vsandvold:
Nobody's complaining, so I'll go a head and merge.

PR #66, opened 2017-02-14
makes name default sorting

### LINKED ISSUE #65: Sort order should have a default
2017-02-13-211825_303x88_scrot

By undefined seems a bit odd! Want to have a look @vsandvold ?

vsandvold:
Closes #65

### DIFF
--- a/src/component/feature/list-container.jsx
+++ b/src/component/feature/list-container.jsx
@@ -21,38 +21,40 @@ const mapStateToProps = (state) => {
         );
     }
 
-    if (settings.sort) {
-        if (settings.sort === 'enabled') {
-            features = features.sort((a, b) => (
-                // eslint-disable-next-line
-                a.enabled === b.enabled ? 0 : a.enabled ? -1 : 1
-            ));
-        } else if (settings.sort === 'created') {
-            features = features.sort((a, b) => (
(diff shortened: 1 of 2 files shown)


src/component/feature/list-container.jsx
```diff
@@ -21,38 +21,40 @@ const mapStateToProps = (state) => {
...
-                if (a.name > b.name) { return 1; }
-                return 0;
-            });
-        } else if (settings.sort === 'strategies') {
-            features = features.sort((a, b) => (
-                a.strategies.length > b.strategies.length ? -1 : 1
-            ));
-        } else if (settings.sort === 'metrics') {
-            const target = settings.showLastHour ? featureMetrics.lastHour : featureMetrics.lastMinute;
+    if (!settings.sort) {
```
vsandvold: Or does it belong in the settings reducer?

```
function getInitState () {
    const feature = {sort: 'name'};

    try {
        const state = JSON.parse(localStorage.getItem(SETTINGS));
        return state ? fromJS(state) : new $Map({ feature });
    } catch (e) {
        return new $Map({ feature });
    }
}
```

ivarconr: no I don't think so, the settings store does not know anything about the content.

PR #74, opened 2017-02-23
Add some tests

SimenB:
This adds some very basic tests using jest, mostly as a PoC.

Builds upon #72, so you can disregard the first commit if you want.


src/component/common/__tests__/util-test.jsx
```diff
@@ -0,0 +1,7 @@
+import { formatFullDateTime } from '../util';
+
+test('formats dates correctly', () => {
+    expect(formatFullDateTime(1487861809466)).toEqual('02/23/2017, 3:56:49 PM');
```
SimenB: node doesn't ship with locales. Might be a stupid test?

SimenB: Have to mock TZ at least, so might be stupid tests?

PR #82, opened 2017-06-28
editable=false on strategies

sveisvei:
* remove delete button for built-in strategies
* remove edit tab for built-in strategies

### DIFF
--- a/src/component/strategies/list-component.jsx
+++ b/src/component/strategies/list-component.jsx
@@ -33,7 +33,11 @@ class StrategiesListComponent extends Component {
                                         <strong>{strategy.name}</strong>
                                     </Link>
                                 </ListItemContent>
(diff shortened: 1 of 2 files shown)


src/component/strategies/list-component.jsx
```diff
@@ -33,7 +33,11 @@ class StrategiesListComponent extends Component {
                                         <strong>{strategy.name}</strong>
                                     </Link>
                                 </ListItemContent>
-                                <IconButton name="delete" onClick={() => removeStrategy(strategy)} />
+                                {
+                                    strategy.editable === false ?
+                                    '' :
```
SimenB: are empty strings stripped out like `null` is?

sveisvei: ListItem parent component did not like null - so I am being pragmatic.

PR #88, opened 2017-07-10
Fix deprecations

SimenB:
This uses the new `prop-types` package, and removes a couple of unused deps.

It also fixes 2 optimzation bailouts caused by using `require` of `exported` stuff


SimenB:
@sveisvei @ivarconr thoughts?

PR #89, opened 2017-07-20
increase slider thumb touch target

vsandvold:
Trying to increase the touch target size as requested in https://github.com/Unleash/unleash/issues/254

Not really sure this works on iOS, though. Needs more work.

### DIFF
--- a/src/component/feature/form/strategy-input-percentage.jsx
+++ b/src/component/feature/form/strategy-input-percentage.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { Slider } from 'react-mdl';
+import styles from './strategy.scss';
 
 const labelStyle = {
     margin: '20px 0',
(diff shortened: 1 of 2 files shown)


ivarconr:
Any chance to settle this @vsandvold ?

vsandvold:
It looks like a really JS-heavy solution is required to get this working on iOS. Styling a larger touch target works like a charm in other browsers, just not in Mobile Safari (which was the use case of the original request). It should be possible using touch and drag event listeners, but I'd rather not go there. Love the idea, but not worth the effort as I see it.

I'll post some links for future reference before closing this PR.

ivarconr:
Aight, thanks for update!

vsandvold:
Another drawback with range inputs in Mobile Safari is that it only respons to touch on the slider thumb and not elsewhere, like the slider bar.

It doesn't feel right to re-implement the slider that comes with Material Design Lite (react-mdl), but here's a few interesting links in case we would like to try something like that later.

Example of range input styling that may produce a larger touch target:
https://github.com/OnsenUI/OnsenUI/pull/2088/files

Example of JS-type solution:
http://rangeslider.js.org/

PR #106, opened 2018-01-29
Move metrics poller to seperate class

### LINKED ISSUE #104: Stop polling when 401 is detected
screenshot from 2018-01-23 09-39-04

ivarconr:
Instead of having various components starting a metrics poller we can move this logic to a specialized class handling an in background poller. We should not fetch metrics before we know that there exists any toggles. 

Todo:
- [x] Write unit tests

This closes #104


src/index.jsx
```diff
@@ -34,6 +35,7 @@ if (process.env.NODE_ENV !== 'production' && window.__REDUX_DEVTOOLS_EXTENSION_C
 }
 
 const unleashStore = createStore(store, composeEnhancers(applyMiddleware(thunkMiddleware)));
+new MetricsPoller(unleashStore);
```
SimenB:
```
  38:1  error  Do not use 'new' for side effects  no-new
```

Move `start` out of the constructor and call it explicitly?

src/metrics-poller.js
```diff
@@ -0,0 +1,36 @@
...
+        this.timer = undefined;
+        this.running = false;
+
+        this.start();
+    }
+
+    start() {
+        const store = this.store;
+        store.subscribe(() => {
+            const features = store.getState().features.toJS();
```
SimenB: is the `toJS()` needed?

src/metrics-poller.js
```diff
@@ -0,0 +1,36 @@
...
+                this.timer = setInterval(this.fetchMetrics.bind(this), 5000);
+            }
+        });
+    }
+
+    fetchMetrics() {
+        this.store.dispatch(fetchFeatureMetrics());
+    }
+
+    destroy() {
```
SimenB: is this ever called? Is it a problem if it's not?

ivarconr: it is via tests. It should not be a problem. Unleash will poll metrics as long as unleash is running and there are active toggles.

PR #110, opened 2018-02-04
Default strategy is actually default. 

ivarconr:
Select the default strategy if the user has not added any strategies when creating new feature toggle. 

Closes https://github.com/Unleash/unleash/issues/204


sveisvei:
IMO this should be solved backend as well.

The "default" is that no special strategies are selected.

ivarconr:
> IMO this should be solved backend as well.

I agree, and if the BE handles it, we would actually end up with just having to change the empty-text, less logic in UI is always better :dancer:

Issue #107, opened 2018-02-02
Timezone should be configurable


ivarconr:
Currently it is hard-coded to Norwegian formatting and timezone. It should be configurable for the user to set whatever timezone he wants, and it should be stored (in local storage or similar).

ivarconr:
for reference see https://github.com/Unleash/unleash-frontend/blob/master/src/component/common/util.js

PR #111, opened 2018-02-07
Timezone

corinnekrych:
closes https://github.com/Unleash/unleash-frontend/issues/107


src/component/common/util.js
```diff
@@ -6,5 +6,25 @@ const dateTimeOptions = {
     minute: '2-digit',
     second: '2-digit',
 };
+// todo for a more comprehensive list use of moment.tz from https://github.com/moment/moment-timezone
+const predefinedLocale = [
+    {
+        locale: 'nb-NO',
+        timezone: 'UTC',
```
ivarconr: Norway is on timezone `Europe/Oslo`

src/component/common/util.js
```diff
@@ -6,5 +6,25 @@ const dateTimeOptions = {
     minute: '2-digit',
     second: '2-digit',
 };
+// todo for a more comprehensive list use of moment.tz from https://github.com/moment/moment-timezone
```
ivarconr: Would be nice to support "all", but how large would that dependency be?

src/component/common/util.js
```diff
@@ -6,5 +6,25 @@ const dateTimeOptions = {
...
+        timezone: 'America/New_York',
+    },
+    {
+        locale: 'en-GB',
+        timezone: 'Europe/London',
+    },
+];
+export const formatFullDateTimeWithLocale = (v, locale) => {
+    let found = predefinedLocale.find(v => v.locale === locale);
+    dateTimeOptions.timeZone = found ? found.timezone : 'UTC';
```
ivarconr: This actually limits the possible time-zones to the defined list. By omitting the timezone we will get the current timezone in the browser, which is probably closer to what the user expects. 

By only changing the locale, and not the timezone, we allow the user to change the formatting. What do you think?

src/component/user/show-user-container.jsx
```diff
@@ -1,13 +1,16 @@
...
+import { updateSettingForGroup } from '../../store/settings/actions';
 
 const mapDispatchToProps = {
     fetchUser,
+    updateSettingLocation: updateSettingForGroup('location'),
 };
 
 const mapStateToProps = state => ({
     profile: state.user.get('profile'),
+    location: state.settings ? state.settings.toJS().location : {},
```
ivarconr: `location` or `locale`?

src/component/feature/metric-component.jsx
```diff
@@ -38,13 +38,16 @@ export default class MetricComponent extends React.Component {
         featureToggle: PropTypes.object.isRequired,
         fetchSeenApps: PropTypes.func.isRequired,
         fetchFeatureMetrics: PropTypes.func.isRequired,
+        location: PropTypes.object,
```
ivarconr: I'm perfectly fine with passing the `locale` details as props as long as it is just a few components that needs it.

**But**; is there a way to set it at root level and just have a simple date-format-function that could be required when needed?

src/component/common/util.js
```diff
@@ -6,5 +6,25 @@ const dateTimeOptions = {
...
+    return new Date(v).toLocaleString(locale, dateTimeOptions);
+};
 
-export const formatFullDateTime = v => new Date(v).toLocaleString('nb-NO', dateTimeOptions);
+export const formatFullDateTime = v => formatFullDateTimeWithLocale(v, 'nb-NO');
```
ivarconr: I guess this would not be needed anymore?

ivarconr:
Thanks for talking the time to dig in to this. Nice tough on the flags! I have tried to give a few inline-feedbacks. 

After looking through your suggestion I have ended up with a few questions:
- Do we really have to specify timezone? Would it not be sufficient to rely on [browser default timezone](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString)? 
- Should we default to browser default locale (navigator.language, provably via [browser-locale](https://github.com/maxogden/browser-locale)), but give the user the possibility to change for himselves?
- ...or would it make sense to have a default locale set up via the `unleash-server`. I guess most companies would have a preferred locale in their environment?
- Is there a simple way to set-up locale handling with React at root-level?

corinnekrych:
@ivarconr Thanks for the review!

The initial purpose of the PR/issue was to be able to change the timezone as a configurable item. Now integrating your suggestions, it’s more a select a configurable locale ;)
Defaulted to browser timezone (ie: not specifying it). Interesting to note date and timezone is not a `chrome://settings` anymore.
Defaulted to navigator locale using `navigator.language` and allow user to change it with a short-listed choices.
Note: we can later have a user profile/setting page with more options:
- set local with a complete list
- set the timezone with a more complete dropdown
All the code is here to make it configurable.

For your comment on root-level setup for the locale state, I think it defats the purpose of `react-reduce`, interesting article here: https://medium.com/@schiehll/setting-global-state-with-react-d702236a3145 but I’d rather stick to react-redux.

ivarconr:
> The initial purpose of the PR/issue was to be able to change the timezone as a configurable item. Now integrating your suggestions, it’s more a select a configurable locale ;)

Nice. I think this is a good first step in making it more configurable. 

> Defaulted to browser timezone (ie: not specifying it). Interesting to note date and timezone is not a chrome://settings anymore.

Wow, I was not aware of this!

> Note: we can later have a user profile/setting page with more options:
> - set local with a complete list
> - set the timezone with a more complete dropdown

Agreed, a user settings page would be a natural place for this in the future.

> For your comment on root-level setup for the locale state, I think it defats the purpose of react-reduce, interesting article here: medium.com/@schiehll/setting-global-state-with-react-d702236a3145 but I’d rather stick to react-redux.

I was thinking something more in direction of [react-intl](https://github.com/yahoo/react-intl), with locale stored in settings store. I am perfectly fine with the current implementation as is. Adding localisation supports will also bring complexity we don't need if we are just talking about formatting a few timestamps.

src/component/user/user.scss
```diff
@@ -2,4 +2,14 @@
...
     border: 2px solid #ffffff;
-}
\ No newline at end of file
+}
+
+.showLocale img {
+    border-radius: 2px;
+    height: 20px;
+    border: 2px solid #ffffff;
+}
```
ivarconr: How about dropping border, increasing height to 30px and adding some margin? 

```css
    border-radius: 2px;
    height: 30px;
    margin: 0 10px;
```

screenshot from 2018-02-08 20-37-43


(PS! I am no UX expert)

ivarconr:
@corinnekrych ready to merge?

corinnekrych:
@ivarconr Yes!

ivarconr:
Live: http://unleash.herokuapp.com

Issue #92, opened 2017-09-07
View for content of archived toggles


ivarconr:
I would like to be able to see how an archived toggle was configured without restoring it to "normal" toggle status. The archived toggle list items aren't clickable right now, but it would be great if they pointed to a (read-only) view of the configuration the toggle had when it was archived.

(from https://github.com/Unleash/unleash/issues/249)

PR #113, opened 2018-02-08
feat(archive): display strategies details for archived features

corinnekrych:
closes https://github.com/Unleash/unleash-frontend/issues/92 

This is a proposal to display archived features with a toggle to show more details (mainly the strategies). @ivarconr wdyt?

I still need to figure out the snapshot test update.


SimenB:
Can you provide a screenshot of how it looks with your changes applied?

ivarconr:
looks good, but maybe it is possible to highlight the strategies a bit more?

screenshot from 2018-02-09 09-09-46

To fix the snapshot-tests:
1. Add all react-mdl components, looks like you are missing `ListItemAction` to [mocks](https://github.com/Unleash/unleash-frontend/blob/master/src/__mocks__/react-mdl.js)
2. Run `yarn run test -- -u` to update all snapshots.

corinnekrych:
@ivarconr ok i can have a look and display the strategies close to what there is in toggles view.
It would be nice to display the parameters value too. wouldn't it?

ivarconr:
> It would be nice to display the parameters value too. wouldn't it?
Yes :+1: 

It would also be handy to see the "createdAt" timestamp, and then the #111 fix will make the formatting awesome! :fire:

src/component/archive/archive-list-component.jsx
```diff
@@ -14,38 +16,117 @@ class ArchiveList extends Component {
     componentDidMount() {
         this.props.fetchArchive();
     }
+    renderStrategyDetail(feature) {
```
ivarconr: **Minor:** You can choose to introduce sub-components, dedicated to render this part, instead of having these `render*()` functions. This will make the `render`-function a bit more readable.

ivarconr:
I merged this because it was already better than what we had and I wanted to include it as part of 3.0.0. Feel free to iterate more on the feature though!

Issue #78, opened 2017-03-09
"Landing page" for toggles should contain info about the toggle


SimenB:
Right now if I click a toggle from the list, the "landing page" is "metrics". To see the value of the toggle, I have to go to either "history" or "edit". I would prefer to either see a new tab with info about the toggle, or this info added above metrics.

![image](https://cloud.githubusercontent.com/assets/1404810/23750103/b3b4c0fa-04cb-11e7-9d26-358c27cf4cea.png)

The only info I get ^ is what strategy it uses. Useful info is hidden in "edit".

![image](https://cloud.githubusercontent.com/assets/1404810/23750134/cdf8cbbe-04cb-11e7-8923-0f288be964c6.png)

I want a read-only view of ^

vsandvold:
I agree, we could try to show the three tabs as sections underneath each other, more like a dashboard. Strategies first perhaps, and then metrics and history last.

PR #116, opened 2018-02-11
Make.strategies.default.tab

corinnekrych:
closes https://github.com/Unleash/unleash-frontend/issues/78

This is an easily-developed proposal for display Feature details information at first sight. 
- Rename `edit` to `strategies`. the fact the field are editable and the `update` buton are enough information to understand you can edit those fields.
- Remove duplicate name/description

wdyt @ivarconr ?


src/component/feature/form/index.jsx
```diff
@@ -35,29 +26,12 @@ class AddFeatureToggleComponent extends Component {
...
-            description,
-            enabled,
-        } = input;
+        const { description, enabled } = input;
         const configuredStrategies = input.strategies || [];
 
         return (
             <form onSubmit={onSubmit(input)}>
                 <section style={{ padding: '16px' }}>
-                    <Textfield
```
ivarconr: This component is also used when creating a toggle. By removing the name field there is no way of defining the name of a toggle when creating a new one.

ivarconr:
I would prefer that you see the metrics and are able to configure activation strategies on the landing page of a feature toggle. 

I think it is fine to have the History as a separate tab.

src/component/feature/form/index.jsx
ivarconr: ...on that note: Having the same component for create and edit might not be the best approach if we end up with a lot of if/else clutter imho.

vsandvold:
This PR touches onto a few ideas we have discussed in the past, but haven't followed through. I really like the what @corrinekrych is doing, and I don't want to impose old ideas on her work. So please just consider the following suggestions as well-meaning advice.

Having a default strategy enabled by default allows us to simplify the create toggle dialog by removing the option to add strategies. Adding/changing strategies will then be deferred to the Strategies tab (renamed from Edit) of the feature toggle view.

The shared component for editing name and description can then be provided by an "Edit" action from a pencil icon in the top right corner of the main feature toggle card. The edit form would then be presented in a popup with clear Save/Cancel actions. I'll try to provide some visual examples tomorrow.

Keep up the good work @corrinekrych :-)

src/component/feature/form/index.jsx
corinnekrych: oops that's bad indeed :)) Let me put it back. Indeed we could do conditional. Let's see how it looks like.

corinnekrych:
@ivarconr we could remove one tab and have only 2: 
- the first tab and default one: the strategies are displayed on the top + the metics at the bottom. We can make the 2 metric s icon smaller. To gain place, we could also display the warning instead of the metrics icons if there is no activities. Ee can rename the tab to: Details or Toggles Details or Configuration
- the second tab to display the history

@vsandvold do we want an edit button. Why not a direct inline edit? We don't have owner right that would need to disable edit for certain users etc.. I'd go for direct edit (simpler UX).

Let me know your thoughts and I can update the PR.

ivarconr:
If we implement the suggestion from @vsandvold  we are only talking about inline-edit possibility for the "description" field (you are not allowed to change the toggle name). And how often do you change the description of a toggle? If it adds complexity to do it inline then we should not bother and have it as a separate view (same as create, but with name field disabled). 

I think it will make a lot of sense to differentiate the process of defining a toggle from adding and updating activation strategies (which also will affect the metrics). :+1:

corinnekrych:
Guys (@sveisvei @ivarconr @vsandvold) , as per your suggestion, I've made description inline-editable, I've moved it out of strategies tab. with this commit #4e871f1
we got:
toggle_edit

ivarconr:
I like it, looks really clean :+1: 

Would you mind adding some unit tests also? The one thing I regret is that we did not add any unit-tests at all for the UI.

src/component/feature/form/form-update-feature-container.jsx
```diff
@@ -34,10 +34,13 @@ const prepare = (methods, dispatch) => {
...
             });
         }
+        delete input.description;
+
         // TODO: should add error handling
         requestUpdateFeatureToggle(input)(dispatch)
             .then(() => methods.clear())
-            .then(() => hashHistory.push(`/features/view/${input.name}`));
+            .then(() => hashHistory.push(`/features`))
+            .then(() => hashHistory.push(`/features/strategies/${input.name}`));
```
ivarconr: This feels a bit weird now, change the url to "/features" and then change it back to the current url. Would it not be better to drop both `hashHistory.push()` calls now?

src/component/feature/form/form-update-feature-container.jsx
```diff
@@ -34,10 +34,13 @@ const prepare = (methods, dispatch) => {
                 delete s.id;
             });
         }
+        delete input.description;
```
ivarconr: Must be dropped if we don't do hashHistory.push

src/component/feature/form/form-update-feature-container.jsx
ivarconr: But maybe there should be some message saying that the toggle was saved successfully? 

We could for example reuse the `Snackbar` used via the error-store.

corinnekrych: let me remove the second line which was required when we used the same container for both update and create.

corinnekrych:
@ivarconr I've added some extra tests. I try to test both the render and the interaction. I made improvement on the coverage but still work to be left to get a good coverage testing. I'll stop here not to make to many file changed in this PR.

ivarconr:
Awesome work! Thanks again @corinnekrych  :fire: :champagne: :+1: :medal_sports:

sveisvei:
👍

Issue #115, opened 2018-02-11
Show history of archived toggles?


ivarconr:
It would be nice to see the change history of a archived toggle. 

What do you think @corinnekrych ?

corinnekrych:
Agreed @ivarconr we could rework the archive page to get it better.

I think we should have a dedicated page to show archived feature (it would fit material design better). It would be nice to reuse the feature display in read-only mode. Out of the 3 tabs: metrics, edit, history, we could display: edit (renamed Strategies) and history. 
If we want to display history, we'll need to fetch the history as the archive endpoint does not return it in the payload.
https://github.com/Unleash/unleash/blob/master/docs/api/admin/feature-toggles-api.md#archive 

Some interesting refactoring is needed for re-using the page.
I can pick up that issue :)

ivarconr:
A dedicated read-only view makes sense. There is already an issue on cleaning up the landing page for active toggles: https://github.com/Unleash/unleash-frontend/issues/78

For the history tab I think the existing component with the current container wrapper will just work as it already uses the events API:

http://unleash.herokuapp.com/api/admin/events/TEST

PR #121, opened 2018-03-11
Archive.revisited

corinnekrych:
fixes https://github.com/Unleash/unleash-frontend/issues/115


corinnekrych:
@ivarconr As discussed in the issue, this PR:
* Make the archive list looks like feature list. 
*Display of archive details is a read-only view.

ivarconr:
Nice! :fire: :+1: :zap: 

I especially enjoy that each toggle reuse the landing page, giving all details about that toggle in read-only mode.

Some thoughts:
- Should the `strategies` section also be displayed as "read-only" (not that there is a save button anyway)
- Could it be made even more highlighted that the toggle is archived? 
- Should you be able to revive a toggle from the details? 
- Minor: Does it make sense to sort archived toggles based on metrics?
- Should it be possible to filter archived toggles via name (aka the search field)?

src/component/archive/archive-list-container.js
```diff
@@ -1,15 +1,80 @@
 import { connect } from 'react-redux';
-import ListComponent from './archive-list-component';
+import ArchiveList from './archive-list-component';
 import { fetchArchive, revive } from './../../store/archive-actions';
+import { updateSettingForGroup } from './../../store/settings/actions';
 
 const mapStateToProps = state => {
-    const archive = state.archive.get('list').toArray();
+    const featureMetrics = state.featureMetrics.toJS();
```
ivarconr: this mostly looks like a copy of `feature/list-container.jsx`. Could we find a way to share that code?

corinnekrych:
> Nice! 🔥 👍 ⚡️
Thanks 😳 
> I especially enjoy that each toggle reuse the landing page, giving all details about that toggle in read-only mode.
>
> Some thoughts:
>
> * Should the strategies section also be displayed as "read-only" (not that there is a save button anyway)
I've just noticed I left the `Add` button, let me remove it. noticed the remove button is not display in this read-only view. 
> * Could it be made even more highlighted that the toggle is archived?
My UX skills are weak here... But should we add a `Chip` or a label somewhere with `Archived`
> * Should you be able to revive a toggle from the details?
Good point, let me add that
> * Minor: Does it make sense to sort archived toggles based on metrics?
Here I reused the same settings('feature') so the list is sorted exactly the same way as features list.
> * Should it be possible to filter archived toggles via name (aka the search field)?
Good point, let's try to reuse the `list-container` too. 

I'LL BE BACK with some code  soon....

corinnekrych:
@ivarconr Here is a second iteration:
screenshot 2018-03-14 11 27 08_preview

I didn't add any `Chip` with _this is an archive_ but with the `Revive` button plus the read-only page, it's clearer we're dealing with archive.

> Note I kept the `revive` button where the `archive` button was, I think having all action buttons grouped at the bottom will make clearer design. we can address in separate issue tho.

ivarconr:
> Note I kept the revive button where the archive button was, I think having all action buttons grouped at the bottom will make clearer design. we can address in separate issue tho.

agreed, let's do that in a separate issue.

src/component/feature/feature-list-item-component.jsx
```diff
@@ -42,27 +43,61 @@ const Feature = ({
...
             </span>
-            <span className={styles.listItemToggle}>
-                <Switch
-                    title={`Toggle ${name}`}
-                    key="left-actions"
-                    onChange={() => toggleFeature(name)}
-                    checked={enabled}
-                />
-            </span>
+            {toggleFeature ? ( // display feature list, toggleFeature is available
```
ivarconr: We end up with a lot of ternary operators here to see if it is a list of archived toggles or active toggles. Would it be better to split it in two components, one for archived and one for active, and split out shared parts out as smaller components?

src/component/feature/form/form-view-feature-component.jsx
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import StrategiesSection from './strategies-section-container';
+import { Button, Icon } from 'react-mdl';
+
+class ViewFeatureComponent extends Component {
+    // static displayName = `UpdateFeatureComponent-{getDisplayName(Component)}`;
```
ivarconr: Curious: does this serve any purpose when commented out?

src/component/feature/form/form-view-feature-component.jsx
```diff
@@ -0,0 +1,42 @@
...
+    componentWillMount() {
+        // TODO unwind this stuff
+        if (this.props.initCallRequired === true) {
```
ivarconr: why do you do this?

src/component/feature/view-component.jsx
```diff
@@ -134,17 +164,35 @@ export default class ViewFeatureToggleComponent extends React.Component {
...
                     <span style={{ paddingRight: '24px' }}>
-                        <Switch
-                            ripple
-                            checked={featureToggle.enabled}
-                            onChange={() => toggleFeature(featureToggle.name)}
-                        >
-                            {featureToggle.enabled ? 'Enabled' : 'Disabled'}
-                        </Switch>
+                        {this.isFeatureView ? (
+                            <Switch
```
ivarconr: could this be written as, instead of using a ternary?

```
<Switch
  disabled={this.isFeatureView}
  ....
</Switch=

corinnekrych:
@ivarconr you're right we can simplify the syntax bu removing most of the ternary `if` in the code. wdyt?

ivarconr:
Sorry, I have been occupied last week and not followed up. I think it looks fine now and in line with the rest of the code base. 

When we start looking in to replacing `react-mdl` we should probably also find better ways to break up components in order to avoid as much of if/else in the rendering as possible. 

I have added you as a collaborator on this project. Feel free to merge yourself when you feel it is ready.

PR #123, opened 2018-04-16
fix(signout): make signout works with proxy

corinnekrych:
Fixed https://github.com/Unleash/unleash-frontend/issues/124


ivarconr:
Nice, did not realize this would cause any issues.

- would it make sense to inform the user about "you are now logged out"? (A landing page or similar)
- would it make sense to have an "LOGOUT_SUCCESS" action and have all stores subscribing to that and clear their state after a successful logout?

PR #137, opened 2018-08-06
React router v4

ivarconr:
Migrating to react-router v4 is a pretty large re-write and it would be nice to have another pair of eyes on it. 

Breadcrumbs and menu generators had to be totally re-written and I have tried to keep all routes inside [component/menu/routes.js](https://github.com/Unleash/unleash-frontend/blob/react-router-v4/src/component/menu/routes.js) to have the routes defined in one place and use that to generate `<Route ..>` components where needed. 

To be fixed before merge:
- [x] Rewrite to react-router 4 (drawer, bottom links, breadcrumb, document.title)
- [x] Isolate all routes
- [x] Add unit tests
- [x] Fix feature create so that it navigates to `/features` after create


src/component/feature/form/form-update-feature-container.jsx
```diff
@@ -42,13 +41,13 @@ const prepare = (methods, dispatch) => {
         // TODO: should add error handling
         requestUpdateFeatureToggle(input)(dispatch)
             .then(() => methods.clear())
-            .then(() => hashHistory.push(`/features`));
+            .then(() => ownProps.history.push(`/features`));
```
ivarconr: I am not sure if this is the right response to an update, but want to keep that in a separate issue, see: https://github.com/Unleash/unleash-frontend/issues/139

src/component/app.jsx
```diff
@@ -1,43 +1,23 @@
...
-    return input;
-}
+import Features from '../page/features';
+import { DrawerMenu } from './menu/drawer';
+import { FooterMenu } from './menu/footer';
+import Breadcrum from './menu/breadcrumb';
+import { routes } from './menu/routes';
 
 export default class App extends Component {
     static propTypes = {
```
SimenB: You should be able to install https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-router-dom_v4.x.x/flow_v0.63.x-/react-router-dom_v4.x.x.js (`flow-typed install react-router-dom@4.x.x`) to get more accurate typing

src/component/menu/__tests__/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
```
SimenB: can use `overrides` in the root eslintrc to avoid this new file. No biggie, though

SimenB:
Thoughts on https://reach.tech/router? Since it's already a rewrite 🙂

src/component/app.jsx
ivarconr: Ahh thanks, I will check it out!

src/component/menu/__tests__/.eslintrc
ivarconr: of course, that will be better :+1:

ivarconr:
[reach.tech/router](https://reach.tech/router) looks pretty similar to `react-router:4` but with a slightly cleaner API :+1:  

I also like the approach with `navigate` where you can import it wherever you need to navigate problematically, and don't have to pass it in as part of properties, as you have with the `history` in react-router-4.

src/component/app.jsx
ivarconr: Because I am using Visual Studio Code I already got the ts definitions for [react-router-dom](https://github.com/DefinitelyTyped/DefinitelyTyped) automatically in the IDE, can't see that these flow types will add much more?

PR #140, opened 2018-08-14
fix(webpack): Upgrade webpack to 4.x

ivarconr:
Seems to be working now. Both the dev mode and production builds looks fine. (TODO: The production builds needs to be verified before this is merged.)

Seems like `webpack-dev-server` should be replaced with `webpack-serve`. However, the proxy part seems to require some extra configuration, and thus I want that in a separate PR.


webpack.config.js
```diff
@@ -3,11 +3,16 @@
 
 const path = require('path');
 const webpack = require('webpack');
-const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const devMode = process.env.NODE_ENV !== 'production';
```
SimenB: `mode` should be set for the config as well

PR #147, opened 2018-08-27
chore(pacakge): Upgrade enzyme to verison 3.5.0

ivarconr:


src/component/feature/form/__tests__/__snapshots__/form-add-feature-component-test.jsx.snap
```diff
@@ -2,7 +2,6 @@
 
 exports[`render the create feature page 1`] = `
 ShallowWrapper {
```
SimenB: these snapshots are weird... Don't we have https://github.com/adriantoine/enzyme-to-json?

ivarconr:
tu @SimenB, looks better now!

PR #159, opened 2018-10-20
Locale fixes

ddunkin:


src/component/user/show-user-component.jsx
```diff
@@ -26,19 +26,15 @@ export default class ShowUserComponent extends React.Component {
     }
 
     updateLocale() {
-        const locale = this.props.location
-            ? this.props.location.locale
-            : this.possibleLocales[this.possibleLocales.length - 1];
+        const locale = (this.props.location && this.props.location.locale) || navigator.language;
```
ivarconr: This will actually not work for IE11 ([see details](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language)), where we have to use `navigator.userLanguage`. 

Would you mind creating a function for extracting the default languages from the browser which first checks `navigator.language` and then `navigator.userLanguage` if that does not exists?

src/component/user/show-user-component.jsx
```diff
@@ -11,7 +11,7 @@ export default class ShowUserComponent extends React.Component {
     };
     possibleLocales = [
         { value: 'nb-NO', image: 'nb-NO' },
-        { value: 'us-US', image: 'us-US' },
+        { value: 'en-US', image: 'en-US' },
```
ivarconr: thanks :+1:

src/component/user/show-user-component.jsx
ddunkin: I made the change, but it doesn't work on IE11 anyway because there are methods that need polyfills, e.g. Array.prototype.find and String.prototype.startsWith.

PR #164, opened 2018-12-19
Add support for permission system in unleash frontend

Kouzukii:
This pull request implements the permission system from Unleash/unleash#373 in the frontend.

Most notable changes are:
* Handle 403 errors from API
* Hide/disable elements in the UI when permissions aren't set


src/component/feature/feature-list-item-component.jsx
```diff
@@ -14,6 +14,7 @@ const Feature = ({
     metricsLastHour = { yes: 0, no: 0, isFallback: true },
     metricsLastMinute = { yes: 0, no: 0, isFallback: true },
     revive,
+    updateable,
```
ivarconr: This one is never set to any value as far as I can see.

src/data/helper.js
```diff
@@ -23,12 +23,25 @@ export class AuthenticationError extends Error {
...
 export function throwIfNotSuccess(response) {
     if (!response.ok) {
         if (response.status === 401) {
             return new Promise((resolve, reject) => {
                 response.json().then(body => reject(new AuthenticationError(response.status, body)));
             });
+        } else if (response.status === 403) {
+            return new Promise((resolve, reject) => {
+                response.json().then(body => reject(new ForbiddenError(response.status, body)));
+            });
```
ivarconr: We have actually used status code `403` in the scenario where the toggle name is already in use. 

This change ends up changing the user feedback to:
image

Previously it was:
image

I think the bug here is that `unleash-server` uses 403 and not `400` for the name-already exists scenario. 

Do you agree? If yes **I will** be happy to change unleash-server to use correct status-code.

src/component/feature/feature-list-item-component.jsx
Kouzukii: I forgot about adding the PermissionComponent here instead of the prop.

src/data/helper.js
Kouzukii: If the permissions system is going to be merged, 403 should be solely reserved for "Forbidden" errors, so for when the client is missing permissions.
Is there any reason you wouldn't be able to change this with a simple number change?

ivarconr: nope. I will head on fixing this ahead of the other PR on unleash-server.

ivarconr: resolved: https://github.com/Unleash/unleash/commit/59ef1d356b9300c276086800e1acbb16df10303c

src/component/common/permission-component.jsx
```diff
@@ -1,47 +1,21 @@
...
-        return deniedCompoinent;
+const PermissionComponent = ({ user, permission, component, otherwise }) => {
+    if (
+        user &&
+        (!user.permissions || user.permissions.indexOf(ADMIN) !== -1 || user.permissions.indexOf(permission) !== -1)
+    ) {
+        return component;
     }
-}
+    return otherwise || '';
```
ivarconr: is it more idiomatically correct to return undefined rather than an empty string?

Kouzukii: React will throw an error if you return undefined or null here

src/component/feature/feature-list-item-component.jsx
```diff
@@ -42,12 +43,26 @@ const Feature = ({
...
+                    }
+                    otherwise={
+                        <Switch
+                            disabled
+                            title={`Toggle ${name}`}
+                            key="left-actions"
+                            onChange={() => toggleFeature(name)}
+                            checked={enabled}
+                        />
+                    }
```
ivarconr: Having a `Permission` component makes it very explicit which is :+1: 

But it feels a bit verbose when you just want to disable a button. Would it make sense to also provide a `hasPermission` function which can be used in these scenarios? 

side:
If we keep the `otherwise` component: Does it not make sense to remove the `onChange` function and add use-feedback on why the component is disabled?

src/component/application/__tests__/application-edit-component-test.js
```diff
@@ -34,7 +34,7 @@ test('renders correctly without permission', () => {
                             {
                                 instanceId: 'instance-1',
                                 clientIp: '123.123.123.123',
-                                lastSeen: 1487861809466,
+                                lastSeen: '2017-02-23T15:56:49',
```
ivarconr: :+1:

PR #166, opened 2019-01-23
Suggestion: Allow loading of custom mdl styles

Kouzukii:
Since we use multiple instances of unleash for different environments, we set them up to use different mdl styles.

This PR would add a backwards compatible way of changing the theme like this:
```js
const {publicFolder, loadStyle} = require('unleash-frontend');
const unleash = require('unleash-server');

loadStyle('deep_orange-pink')
  .then(() => unleash.start({
    publicFolder,
    databaseUrl: DB_URL
  }))
  .catch(err => console.log(err));
```

screenshot 2019-01-23 at 16 53 02

This is just a suggestion, maybe a more elegant solution is possible.

Cons:
* It modifies the module inplace, the theme would stay locked in until the next time loadStyles() is called or the module is reinstalled. So a start without loadStyles() would not start with the blue theme.
* It makes react-mdl a prod-dependency rather than a dev-dependency.


ivarconr:
I like the idea to allow users to change the theme :+1: 

It would however be nice if we could find a way to solve this without requiring react-mdl as a dependency bleeding out, (I also have a plan to replace react-mdl some day, and https://material-ui.com/ looks promising.

ivarconr:
Will be solved via #170

PR #170, opened 2019-03-11
feat: Customisable UI via config

ivarconr:
This feature enables overrides of certain UI elements from the API, such as setting a different background color for the header. This could also solve #166 if we add a theme option. 

The idea is to make it easier to customise the UI, both for different environments, but also to provide company specific overrides (e.g. links to internal setup guides, etc) 

/cc @Kouzukii

### DIFF
--- a/src/store/ui-config/index.js
+++ b/src/store/ui-config/index.js
@@ -1,0 +1,41 @@
+import { Map as $Map } from 'immutable';
+import { RECEIVE_CONFIG } from './actions';
+
+const localStorage = window.localStorage || {
+    setItem: () => {},
+    getItem: () => {},
+};
+
+const basePath = location ? location.pathname : '/';
+const UI_CONFIG = `${basePath}:ui_config`;
+
+const DEFAULT = new $Map({
+    headerBackground: undefined,
+    environment: undefined,
+    slogan: 'A product originally created by FINN.no.',
+});
+
+function getInitState() {
+    try {
+        const state = JSON.parse(localStorage.getItem(UI_CONFIG));
+        return state ? DEFAULT.merge(state) : DEFAULT;
+    } catch (e) {
+        return DEFAULT;
+    }
+}
+
+function updateConfig(state, config) {
+    localStorage.setItem(UI_CONFIG, JSON.stringify(config));
+    return state.merge(config);
+}
+
+const strategies = (state = getInitState(), action) => {
+    switch (action.type) {
+        case RECEIVE_CONFIG:
+            return updateConfig(state, action.value);
+        default:
+            return state;
+    }
+};
+
+export default strategies;
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ before_install: yarn global add greenkeeper-lockfile@1
 before_script: greenkeeper-lockfile-update
 script: yarn run test:ci
 after_script: greenkeeper-lockfile-upload
+cache:
+    yarn: true
 env:
   global:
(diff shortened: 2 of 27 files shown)


ivarconr:
The biggest drawback is that one actually have to do the in initial fetch; It's kind of annoying when the topbar/theme changes color after render. This could maybe be better cached in the client in local-storage or similar.

ivarconr:
So instead of exposing mdl-react themes directly this feature now allows the user to choose from four ui themes. This allows us to move to a different UI framework without tiding the themes to react-mdl.

The desired theme should be specified in unleash-server at startup, making it easy to have different theme in different environment. 

- default (blue_grey-light_blue)
- yellow (amber-light_blue )
- grey (grey-light_blue )
- green (teal-light_blue)

ivarconr:
default: 
image

yellow:
image

grey:
image

green:
image

Kouzukii:
> The biggest drawback is that one actually have to do the in initial fetch; It's kind of annoying when the topbar/theme changes color after render. This could maybe be better cached in the client in local-storage or similar.

You could specify %PUBLIC_URL%/api/config.js in the HTML `<head>` so it's loaded before the site is rendered.

ivarconr:
> You could specify %PUBLIC_URL%/api/config.js in the HTML `<head>` so it's loaded before the site is rendered.

What do you mean?

src/theme-handler.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    setTheme() {
+        const theme = this.store.getState().uiConfig.get('theme');
+        if (this.theme !== theme && isValidTheme(theme)) {
+            this.theme = theme;
+            const themeEl = document.getElementById('theme');
+            setTimeout(() => {
+                themeEl.href = `public/theme/${theme}.css`;
+            }, 90);
```
ivarconr: The timeout here was made to make it not flicker.

ivarconr:
Wait, when I think about it, I could just do the whole select correct theme on the server side as a handling of the "index.html" and just replace link tag directly.

Kouzukii:
> > You could specify %PUBLIC_URL%/api/config.js in the HTML `<head>` so it's loaded before the site is rendered.
> 
> What do you mean?

In addition to `/api/config` which delivers the config as json, the backend will expose `/api/config.js`. 
This endpoint returns a piece of JavaScript code that sets the global variable `window.unleashConfig` so on the inital render of the site the theme and headerBackground is already loaded.

ivarconr:
> > > You could specify %PUBLIC_URL%/api/config.js in the HTML `<head>` so it's loaded before the site is rendered.
> > 
> > 
> > What do you mean?
> 
> In addition to `/api/config` which delivers the config as json, the backend will expose `/api/config.js`.
> This endpoint returns a piece of JavaScript code that sets the global variable `window.unleashConfig` so on the inital render of the site the theme and headerBackground is already loaded.

Yes this will work too.

ivarconr:
What if the backend just inserts the UI config directly to the markup as a global variable in `<head>`? Then everything will just work? No need to expose an API before the data is actually stored in the db serverside.

src/store/ui-config/index.js
```diff
@@ -0,0 +1,38 @@
+import { Map as $Map } from 'immutable';
+import { RECEIVE_CONFIG } from './actions';
+
+// TODO: provde a mock if localstorage does not exists?
+const localStorage = window.localStorage || {};
```
Kouzukii: is supported by all browsers, even IE8, I don't think it's necessary

ivarconr:
> What if the backend just inserts the UI config directly to the markup as a global variable in <head>? Then everything will just work? No need to expose an API before the data is actually stored in the db server side.

After testing: this will still create a flicker as it already has loaded default.css. Ways to avoid flicker:
- have server side replacing the css href in index.html
- block rendering until we have loaded new theme.css

ivarconr:
Because we really need to stop using react-mdl I have decided to not add theme support, as this may need to be adapted to the next ui-framework.

PR #188, opened 2019-09-18
fix: Prevent text highlighting overlap between chips

jonovs:
**Issue**
When double-clicking on a chip, text belonging to a different chip also gets highlighted. After copying and pasting, any unwanted text then has to be removed manually.

**Affected browsers**
Only tested on Chrome 76

**Solution**
The proposed solution is to wrap chip span elements in a flex enabled div which will prevent multiple chip texts from getting highlighted after double-clicking. 

*Before*
Screen Shot 2019-09-18 at 10 20 31 AM
Text selected after double-clicking the chip text: `cancel456`
The "cancel" bit comes from the close button description located in the previous chip in the list.

*After*
Screen Shot 2019-09-18 at 10 24 16 AM
Text selected after double-clicking the chip text: `456`

### DIFF
--- a/src/component/feature/form/__tests__/__snapshots__/strategy-input-list-test.jsx.snap
+++ b/src/component/feature/form/__tests__/__snapshots__/strategy-input-list-test.jsx.snap
@@ -5,24 +5,32 @@ exports[`renders correctly when disabled 1`] = `
   <p>
     featureName
   </p>
-  <react-mdl-Chip
+  <div
     style={
       Object {
-        "marginRight": "3px",
+        "display": "flex",
       }
     }
   >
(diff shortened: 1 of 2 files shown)


jonovs:
`display: flex` is what I found to have worked. Feel free to propose other solutions.

ivarconr:
Nice. Looks good.

jonovs:
@ivarconr It looks like this change breaks the layout if chips span over multiple lines. A quick fix would be to revert this change.

ivarconr:
Bah. Would you mind providing the fix?

jonovs:
See #194

Issue #151, opened 2018-08-28
Switch to "babel-preset-env"


ivarconr:
`babel-preset-es2015`, `babel-preset-stage2` are now deprecated. should use [babel-preset-env](https://babeljs.io/docs/en/babel-preset-env) instead.

luanraithz:
@ivarconr can you close this issue?

PR #190, opened 2019-10-05
fix: babel-preset-env

luanraithz:
#151 
Removed deprecated babel config
Runned link fix


ivarconr:
Thanks for looking at this. I'm a bit busy the next couple of days but will look in to it soon.

Do you understand why one of the snapshot tests fail?

luanraithz:
Sorry, I didn't see this failing test, the problem seems to happen because the [updateConfig function](https://github.com/luanraithz/unleash-frontend/blob/master/src/store/ui-config/index.js#L28), which updates local storage, so the next test will have something [here](https://github.com/luanraithz/unleash-frontend/blob/master/src/store/ui-config/index.js#L20).
If you run only the `should only update headerBackground` test, it works (at least on my machine).
I just don't know if it was a breaking change in the [new jest version](https://github.com/luanraithz/unleash-frontend/blob/master/package.json#L65) or something else.

What should I do? Add a beforeEach to clear localstorage?

ivarconr:
Ahh, good catch. UI options is stored locally as they seldom changes and thus is stupid to must fetch them from server. 

> Add a beforeEach to clear localstorage?

Sounds like a good approach.

luanraithz:
@ivarconr did it

PR #200, opened 2019-12-31
allow hiding of feature toggle

TobyGWilliams:
Hello, 

Love unleash and looking forwards deploying it. However, I find myself activating and deactivating the toggles when trying to navigate to the feature. 

This PR gives the option to pass a flag in the config to hide it.

_Note: This is more of a proposal, the empty span isn't good but the UI library really didn't like being  given  null/undefined.
I'd also like to have written some more specific unit tests instead of relying on the snapshot however I kept running into a memory leak._


ivarconr:
Thanks for this suggestion. This is valuable input as ee are in the process of planning a brand new frontend-implementation as this project uses a deprecated UI framework. 

I was thinking this might be something that should be configurable on a per user basis instead of the whole instance.

PR #203, opened 2020-02-22
style: stick footer at the bottom of the page

sumn2u:
The footer link is not placed at the bottom of the page and it looks not so good. So I create this PR to stick the footer at the bottom of the page.

Happy coding!

### DIFF
--- a/src/component/app.jsx
+++ b/src/component/app.jsx
@@ -42,11 +42,11 @@ export default class App extends PureComponent {
                                 <ErrorContainer />
                             </Cell>
                         </Grid>
-                        <Footer size="mega">
-                            <FooterMenu />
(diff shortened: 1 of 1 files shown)


ivarconr:
I am not convinced making it fixed is the way to go. Now content will flow over it and parts of the footer is hidden. 

image


This is generally only a problem when there is little "main content" in a view. I think the solution is to make sure that the main content fills at least some portion of the screen.

sumn2u:
@ivarconr Thanks for the clarification. I see that we were putting `Footer` inside the `Content ` section so it was causing the problem. 
Now the footer content should be at the bottom of the page :)

ivarconr:
the footer is not showing anymore now for me.

PR #211, opened 2020-05-14
Transform username/password login response to json

goober:
The profile information is not updated correctly when you login with a username and password.


ivarconr:
Will merge and release soon. 

Just need to look in to if we should / could we reset all stores as a response to a fresh login. Suggestions?

goober:
I guess that it could be a good idea to reset the store on a fresh login. Especially if you are switching between users with a different set of permissions

PR #235, opened 2021-01-21
Basic implementation of deprecation/reactivation of strategies

chriswk:
This is not done yet, needs some work on the layout side. The functionality is in place though.


ivarconr:
I tweaked the UI a bit, let me know what you think @chriswk 

image

PS! Pay attention to also disabling the possibility to deprecate the "default" strategy. What do you think?

chriswk:
I like it, is it perhaps something we should validate server side as well? Return 403 for post calls to deprecate endpoint for `default`?

chriswk:
post calls to deprecate endpoint for `default` strategy will now return 403 forbidden, so it's not just blocked here in the GUI, but also at the API level

ivarconr (approved):

PR #236, opened 2021-01-27
feat: Addon support from UI

ivarconr:


chriswk (approved):
Having seen this as a demo, I'm all LGTM :)

PR #243, opened 2021-02-17
Feature/stale dashboard

FredrikOseberg:
Implements the first version of a technical dept dashboard that helps you keep an overview which toggles have outlived their expected lifetime. 

Enhancements left for later:
- Keep track of each individual headers ascending/descending state
- Mobile view
- Bulk actions on toggles
    - Mark toggles as stale
    - Delete toggles
    - Delete toggle confirmation screen
- Loader
- Empty toggle states


package.json
```diff
@@ -38,7 +38,12 @@
     "prepublish": "npm run build"
   },
   "main": "./index.js",
-  "dependencies": {},
+  "dependencies": {
```
FredrikOseberg, line 41: Move to DevDependencies

src/component/menu/__tests__/routes-test.jsx
```diff
@@ -1,12 +1,12 @@
...
 
 test('returns all defined routes', () => {
-    expect(routes.length).toEqual(32);
+    expect(routes.length).toEqual(33);
     expect(routes).toMatchSnapshot();
 });
 
 test('returns all baseRoutes', () => {
-    expect(baseRoutes.length).toEqual(10);
+    expect(baseRoutes.length).toEqual(11);
```
ivarconr, line 9: Not relevant for this PR: 
it's kind of stupid that we have these test, verifying we have managed to put something inside an array. We should probably instead verify that we have critical baseRoutes still in place (e.g. /features, /strategies etc).

ivarconr (approved):
Looks really good. Also had an VC-walk-though. 

Only need to update the dependencies and then this is good to go.

PR #247, opened 2021-02-26
Feat/file structure

FredrikOseberg:
Here is a suggested filestructure for the frontend project.

The changes made includes:
- Match filename with React component name
- Move components/containers and related files into separate folders matching the component name
- Mirror component placement in the file tree with it's place in the react tree
- Localize CSS where it makes sense.
- Match CSS with component name when CSS module only applies for that component.

The benefit of this approach is:
- We can easily delete components and all of their associated files that only are used in that component.
- We can more easily focus on one component at the time, reducing clutter.
- We can easily search for components using editor text search
- Reduce clutter in top level folder and easily refactor large components to contain without ending up with a huge flat structure directory
- We can more easily refactor by reducing mental overhead and having the ability to complete delete anything related to a component or sub component.  

Drawbacks: 
- Filepath imports outside of the component sometimes needs to drop multiple levels down
- More directory nesting compared to a flat structure

As another topic of discussion, most of our components in the page folder seems to be just importing a component and exporting a render function that just renders this component. Is that useful? Perhaps strictly as a mental shift between components and pages, but the components themselves could just as well be directly rendered into the Route mostly.

As an alternative, the page component could take on the responsibility of composition (sort of like the top level component), that would give it some more meaning compared to the current state. 

Thoughts?


FredrikOseberg:
Merged into #250

PR #248, opened 2021-03-01
fix: metrics invalid date

FredrikOseberg:
Fixes an issue where createdAt field in the Metrics component showed: "Invalid date". The following steps have been done to prevent that from happening: 

- Added capabilities in unleash server to return the feature on create
- Added capabilities in unleash-frontend to receive the created feature and insert it into redux store.

### DIFF
--- a/src/component/feature/view/metric-component.jsx
+++ b/src/component/feature/view/metric-component.jsx
@@ -6,6 +6,7 @@ import { Link } from 'react-router-dom';
 import { AppsLinkList, calc } from '../../common';
 import { formatFullDateTimeWithLocale } from '../../common/util';
 import styles from './metric.module.scss';
+import ConditionallyRender from '../../common/conditionally-render';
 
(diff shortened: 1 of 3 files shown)


ivarconr:
> When creating the feature, update state locally then initiate a remote fetch from the server.

I would actually argue that Unleash should return the full feature toggle definition after the create. Then you will avoid the need to do a new fetch the update after create, 


There are certain fields the clients should not be allowed to define (createdAt, lastSeenAt, etc).

FredrikOseberg:
I agree. That is a better solve.

man. 1. mar. 2021 kl. 22:08 skrev Ivar Conradi Østhus <
notifications@github.com>:

FredrikOseberg:
@ivarconr Updated this PR to reflect the fact that unleash server now returns the created feature toggle when calling the endpoint.

ivarconr (approved):

PR #264, opened 2021-04-07
Fix/material UI cleanup

FredrikOseberg:
- General UI/Improvments
- Implemented protected routes and capabilities for standalone routes
- Refactoring
- Began implementing interfaces for users and routes

Later iterations will involve:
- Moving some of this logic serverside, especially ordering of strategies.
- Reset loading counter on logout
- Avoid flashing features page when logging out


src/component/history/EventHistory/EventHistory.jsx
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from 'react';
```
chriswk, line 1: Seems like a candidate for tsx? New file and all?

chriswk (approved):
I see you've moved a couple of components to more semantically correct paths. I like it.

ivarconr:
Not showing "placeholder" for feature toggles when signing in:

https://www.loom.com/share/5780b2497f9241de9518eb874a56e0c4

ivarconr (commented):
All in all this looks like a great improvement. I do not have chance to verify all small fixes in here, but my impression is that this is a huge step in closer to v4.

src/component/user/Login/Login.jsx
```diff
@@ -0,0 +1,58 @@
+import { useEffect } from 'react';
+import classnames from 'classnames';
+import useMediaQuery from '@material-ui/core/useMediaQuery';
+import { useTheme } from '@material-ui/core/styles';
+import { Typography } from '@material-ui/core';
+
+import AuthenticationContainer from '../authentication-container';
+import ConditionallyRender from '../../common/ConditionallyRender';
+
+import { ReactComponent as SwitchesSVG } from '../../../icons/switches.svg';
```
ivarconr, line 10: Could we use the Unleash Logo instead, possible an inverted version to make it work with the dark background?

src/component/history/EventHistory/EventHistory.jsx
FredrikOseberg: Certainly.

src/component/user/Login/Login.jsx
FredrikOseberg: Yep. I'll add that.

src/component/history/EventHistory/EventHistory.jsx
FredrikOseberg: Certainly. But I noticed when I started converting App.tsx to type

> Not showing "placeholder" for feature toggles when signing in:
> 
> https://www.loom.com/share/5780b2497f9241de9518eb874a56e0c4

Thanks. Handling that by resetting loading count on logout now.

FredrikOseberg:
> Not showing "placeholder" for feature toggles when signing in:
> 
> https://www.loom.com/share/5780b2497f9241de9518eb874a56e0c4

Thanks. Handling that now by resetting loading count on logout.

PR #269, opened 2021-04-16
wip: frontend should understand rbac permissions

ivarconr:


src/component/Access/access-context.js
```diff
@@ -0,0 +1,27 @@
...
+const AccessContext = React.createContext()
+
+const AccessProvider = ({store, children}) => {
+  const hasAccess = (permission, project) => {
+    const permissions = store.getState().user.get('permissions');
+
+    return permissions.some(p => {
+      if(p.permission === 'ADMIN') {
+        return true
+      }
```
FredrikOseberg, line 12: Does this method need a check to match permission and return true if project is undefined?

FredrikOseberg:
> 40+ files to go, but does this make sense?

I think so. The alternative is to expose the data through redux and provide as input to a helper. Or we could create a reusable hook. The latter is perhaps more elegant but requires each component where it is used to be a functional component.

src/component/Access/access-context.js
ivarconr: If the users has the ADMIN permission we do not need to check the project as it will take precedence.

FredrikOseberg: Yes. But a user with the Editor role should be able to create and edit addons (at least according to the current description). The way I'm reading this is that we return early if they have ADMIN privileges, and if not we do a check that matches the incoming permission with the user permission and permission project with incoming project. Since addons do not have projects, aren't we going to miss this use case here? Or am I missing something?

ivarconr: Yes you are. The ADMIN permission is a special permission only granted if you belong to the admin role. This permission automatically wins over all other permissions. 

E.g. You need the ADMIN permission to manage users. We might introduce a special permission later for CREATE/UPDATE/DELETE users, but we have not yet. 

Having the ADMIN permission also simplifies the access to all project resources, as this permission will not require you to have a specific permission for for the given project.

src/component/Access/access-context.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+const AccessContext = React.createContext()
+
+const AccessProvider = ({store, children}) => {
```
FredrikOseberg, line 5: Could we separate out the AccessProvider into an AccessProvider.tsx file? It would be in line with the naming convention we've gone for with other React components and make it easier to search for. 

I've also been trying to keep all new files in typescript to reduce the workload further down the road. This isn't a must if you feel it will add significant more time to the PR, but nice to keep in the back of the mind when we create new files going forward.

src/page/admin/users/UsersList/UsersList.jsx
```diff
@@ -83,7 +84,7 @@ function UsersList({
                         <TableCell>Name</TableCell>
                         <TableCell>Username</TableCell>
                         <TableCell align="center">Role</TableCell>
-                        <TableCell align="right">{hasPermission('ADMIN') ? 'Action' : ''}</TableCell>
+                        <TableCell align="right">{hasAccess('ADMIN') ? 'Action' : ''}</TableCell>
```
FredrikOseberg, line 87: Why not use the ADMIN constant here?

FredrikOseberg (approved):
LGTM. Just some minor comments regarding naming conventions and constants. Great job.

PR #277, opened 2021-04-27
fix/locale

### LINKED ISSUE #265: Detected Locale is Broken in Safari
In Safari, `navigator.language` returns lowercase values, e.g. `en-us`, which doesn't match the `en-US` local, leading to an unknown locale.

FredrikOseberg:
Fixes #265


ivarconr:
In future we should probably try to find a list of all legal locales or just choose a few official that makes sense. 

eg.g:
- https://stackoverflow.com/a/63101737

ivarconr (approved):

PR #279, opened 2021-04-28
Fix/bugfixes

FredrikOseberg:


src/component/context/form-context-component.jsx
```diff
@@ -42,26 +68,44 @@ class AddContextComponent extends Component {
...
+        if (editMode) return true;
+
         try {
             await validateName(name);
             errors.name = undefined;
         } catch (err) {
             errors.name = err.message;
         }
-
+        console.log(errors.name);
```
chriswk, line 81:
```suggestion
```

chriswk: I'm assuming we don't want to log the errors to console in production :)

chriswk (approved):

PR #281, opened 2021-04-29
Feat/bootstrap

FredrikOseberg:


src/component/AccessProvider/AccessProvider.tsx
```diff
@@ -1,38 +1,42 @@
...
-}
-
-export default AccessProvider;
\ No newline at end of file
+const AccessProvider: FC<IAccessProvider> = ({ store, children }) => {
+    const hasAccess = (permission: string, project: string) => {
+        const permissions = store.getState().user.get('permissions') || [];
+
+        const result = permissions.some((p: IPermission) => {
+            if (p.permission === ADMIN) {
```
chriswk, line 21: `return p.permission === ADMIN || (p.permission === permission && p.project === project)`
Still not sure that would be easier to read though

src/component/common/ProtectedRoute/ProtectedRoute.jsx
```diff
@@ -11,7 +11,7 @@ const ProtectedRoute = ({
             {...rest}
             render={props => {
                 if (unauthorized) {
-                    return <Redirect to="/login" />;
+                    return <Redirect to={'/login'} />;
```
chriswk, line 14: I guess this makes it relative?

chriswk (approved):
Nifty, seems like this was more complicated than what we thought at the beginning.

src/component/common/ProtectedRoute/ProtectedRoute.jsx
FredrikOseberg: No, but it works because of the basename set in the router: 

```
 <Router basename={`${getBasePath()}`}>
   ... children
</Router>
```

Due to this property our internal routing will be automatically handled by react-router-dom (prepending the basename as the subpath) as long as we use their components to set up and link to routes, leaving the API paths for us to handle.

PR #286, opened 2021-05-05
chore: fix releases URL in changelog

theneva:
The link is broken as the reference `releases-url` doesn't exist, so I replaced it with what I think it's supposed to be.


FredrikOseberg (approved):
Thanks! This looks good to me.

PR #288, opened 2021-05-05
Fix/jumping screen

FredrikOseberg:
- Fixes an issue where opening modals causes the page to jump
- Fixes an issue where changing tabs scrolls to the top of the page


ivarconr (approved):
nice one!

PR #289, opened 2021-05-06
Fix/strategy constraints

FredrikOseberg:


chriswk (approved):
Does this still keep most of the config from create-react-app, if so I'm fine with it,

FredrikOseberg:
> Does this still keep most of the config from create-react-app, if so I'm fine with it,

Yes. It uses [craco](https://github.com/gsoft-inc/craco) to override specific parts of CRA without ejecting.

ivarconr (approved):
LGTM. 

Could you add some details to the readme on how to enable/disable "wdyr"?

PR #293, opened 2021-05-10
fix: proper error handling for auth-settings being stored

ivarconr:


src/page/admin/auth/oidc-auth-container.js
```diff
@@ -0,0 +1,12 @@
+import { connect } from 'react-redux';
+import OidcAuth from './oidc-auth';
+import { getOidcConfig, updateOidcConfig } from '../../../store/e-admin-auth/actions';
+
+const mapStateToProps = state => ({
+    config: state.authAdmin.get('oidc'),
+    unleashUrl: state.uiConfig.toJS().unleashUrl,
+});
+
+const Container = connect(mapStateToProps, { getOidcConfig, updateOidcConfig })(OidcAuth);
```
FredrikOseberg, line 10: This works. But giving a more explicit name is better for auto importers. When we have multiple components named Container my auto importer struggles to figure out which container to import. 

```suggestion
const OidcContainer = connect(mapStateToProps, { getOidcConfig, updateOidcConfig })(OidcAuth);
```

src/page/admin/auth/oidc-auth-container.js
```diff
@@ -0,0 +1,12 @@
...
+
+export default Container;
```
FredrikOseberg, line 12:
```suggestion
export default OidcContainer;
```

FredrikOseberg (approved):
LGTM. Two things could be considered, but aren't showstoppers: 
- New files could be written in TS to reduce load down the road.
- New files could follow new filename naming convention that we are trying to conform to gradually.

FredrikOseberg:
> > * New files could be written in TS to reduce load down the road.
> > * New files could follow new filename naming convention that we are trying to conform to gradually.
> 
> In general I agree. It was a matter of speed in this scenario, and this would have taken me a bit more brain power to get that right! :/

Fair enough :) 

Much will be revised in the future UI/UX upgrade anyways, so it might not be worth right now either way.

PR #304, opened 2021-06-02
Fix/customer journey patches

FredrikOseberg:
Makes improvements to standalone screens by adding the unleash logo, increasing the padding and changing the background color. In addition: 
- Handle a bug where toast is not displayed on first render
- Add more spacing between headers on Add Strategy modal
- Sort roles so Admin, Editor and Viewer is the order they are viewed in the list when adding new user.

Skjermbilde 2021-06-03 kl  15 57 38


chriswk (approved):
Seems like a reasonable move :)

PR #305, opened 2021-06-02
fix: improve passwordchecker

FredrikOseberg:
Handles an obscure case in the password checker where password is not accepted because of three repeating characters. Now the user will receive a warning that three characters in a row is not permitted.

Skjermbilde 2021-06-03 kl  15 59 16


chriswk (approved):
LGTM :)

PR #309, opened 2021-06-19
support offline mode, support icons and fonts

aneeshrelan:
- [x] Moved all usage of `<Icon>{icon}</Icon>` to dedicated icons from material-ui. Except the usage of dynamic application icons
- [x] Included Robot font within app.css
- [x] Validated on an offline mode with no internet

Resolves [#863](https://github.com/Unleash/unleash/issues/863)


public/index.html
```diff
@@ -7,16 +7,7 @@
...
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <meta name="description" content="unleash" />
-
         <title>Unleash - Enterprise ready feature toggles</title>
-        <link
-            href="https://fonts.googleapis.com/icon?family=Material+Icons"
-            rel="stylesheet"
-        />
-        <link
-            href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700"
```
FredrikOseberg, line 17: We need to include the fonts for the different weight types as well (300, 400, 500, 700) and add the correct font weight mapping in the @font-face ruleset. Right our fonts have only the default regular weight regardless of the CSS settings.

src/app.css
```diff
@@ -1,3 +1,7 @@
+@font-face {
+    font-family: 'Roboto';
+    src: url('./assets/fonts/Roboto-Regular.ttf');
+}
```
FredrikOseberg, line 4: The font weight should be included here, as well as the other fonts.

src/component/addons/AddonList/ConfiguredAddons/ConfiguredAddons.jsx
```diff
@@ -62,11 +63,7 @@ const ConfiguredAddons = ({
                             }
                             onClick={() => toggleAddon(addon)}
                         >
-                            <Icon>
-                                {addon.enabled
-                                    ? 'visibility'
-                                    : 'visibility_off'}
-                            </Icon>
+                            {addon.enabled ? <Visibility /> : <VisibilityOff />}
```
FredrikOseberg, line 66:
```suggestion
                           <ConditionallyRender 
                                  condition={addon.enabled} 
                                  show={<Visibility />} 
                                  elseShow={<VisibilityOff />}
                            />
```

src/component/menu/drawer.jsx
```diff
@@ -17,8 +17,8 @@ const filterByFlags = flags => r => {
     return true;
 };
 
-function getIcon(name) {
-    if (name === 'c_github') {
+function getIcon(IconComponent) {
```
FredrikOseberg, line 20: The "library_books" icon is actually coming from our API, which will override the default setting you are setting in the reducer further down and render `<library_books>` into the DOM causing a warning. We need to add a check here for "library_books" like we are doing with the github icon.

src/store/ui-config/index.js
```diff
@@ -21,7 +22,7 @@ const DEFAULT = new $Map({
     links: [
         {
             value: 'Documentation',
-            icon: 'library_books',
+            icon: LibraryBooks,
```
FredrikOseberg, line 25: This is coming dynamically from the API and will override these defaults. See the comment above.

FredrikOseberg (changes requested):
Fantastic start. I added a few comments that we need to resolve before merging, but this is really close. 

If you can open a PR from a branch in your repository it would be easier for me to check out locally.

aneeshrelan (commented):
> Fantastic start. I added a few comments that we need to resolve before merging, but this is really close.
> 
> If you can open a PR from a branch in your repository it would be easier for me to check out locally.

Thanks for the initial feedback. I'll resolve in a couple of days, if that's okay and probably re-create the PR to change the source branch

public/index.html
```diff
@@ -7,16 +7,7 @@
         <meta name="baseUriPath" content="::baseUriPath::" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <meta name="description" content="unleash" />
-
         <title>Unleash - Enterprise ready feature toggles</title>
-        <link
```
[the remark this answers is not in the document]
FredrikOseberg: Yes you're correct. We'll keep it in place for now.

PR #311, opened 2021-06-28
fix: add logout as an explicit call

ivarconr:
I have not removed the LogoutComponent and the logout logic from the stores. Would we like to keep that around, or should I just remove it all?


src/component/menu/drawer.jsx
```diff
@@ -99,6 +100,10 @@ export const DrawerMenu = ({
                         {item.title}
                     </NavLink>
                 ))}
+                <a className={classnames(styles.navigationLink)} href={`${getBasePath()}/logout`}>
+                    {getIcon('exit_to_app')}
```
FredrikOseberg, line 104: Just merged a PR on the offline capabilities. With these changes, this needs to be changed to: 

```
import IconName from 'material-ui/core'

getIcon(IconName)
```

FredrikOseberg (changes requested):
Looks good, due to some changes that were done in the meanwhile we need to change how we retrieve the icon.

FredrikOseberg:
> I have not removed the LogoutComponent and the logout logic from the stores. Would we like to keep that around, or should I just remove it all?

I'd like to get rid of it, but we are currently using user.authDetails to determine whether or not a user is authenticated in the frontend. If we no longer update this on logout, we need another method of determining user authentication.

ivarconr:
> If we no longer update this on logout, we need another method of determining user authentication.

we kind of do because we instructs the browser to clear all cookies and local storage, so the authDetails will be gone after the server-side redirect.

src/component/menu/__tests__/routes-test.jsx
```diff
@@ -1,7 +1,7 @@
 import { baseRoutes, getRoute } from '../routes';
 
 test('returns all baseRoutes', () => {
-    expect(baseRoutes.length).toEqual(12);
+    expect(baseRoutes.length).toEqual(11);
```
chriswk, line 4:
```suggestion
    expect(baseRoutes).toHaveLength(11);
```

PR #321, opened 2021-08-10
fix: loadNameFromHash pull name search param

shanestillwell:
* The current code is looking for the name in the hash, but the name is
  not passed in the hash, but the search (queryString) parameter.
* Uses the [URLSearchParams](http://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
  browser API to find the name
* Maybe this function should be renamed to `loadNameFromSearch`?

Fixes https://github.com/Unleash/unleash/issues/886

### DIFF
--- a/src/component/common/util.js
+++ b/src/component/common/util.js
@@ -85,14 +85,9 @@ export function updateWeight(variants, totalWeight) {
     });
 }
 
-export function loadNameFromHash() {
-    let field = '';
-    try {
-        [, field] = document.location.hash.match(/name=([a-z0-9-_.]+)/i);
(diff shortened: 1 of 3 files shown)


ivarconr:
Thanks for taking the time to digging in. This was introduced as a result of moving to browser history API and dropping the hash bang. 

> Maybe this function should be renamed to loadNameFromSearch?

Yes please!

FredrikOseberg:
Merged in https://github.com/Unleash/unleash-frontend/pull/323. 

Thanks @shanestillwell!

PR #326, opened 2021-08-23
feat: SSO auto-create users with default role

ivarconr:


src/page/admin/auth/AutoCreate/AutoCreateForm.jsx
```diff
@@ -0,0 +1,96 @@
...
+                    rows={2}
+                    variant="outlined"
+                    size="small"
+                />
+            </Grid>
+        </Grid>
+    </Fragment>);
+}
+
+AutoCreateForm.propTypes = {
```
FredrikOseberg, line 91: Consider creating new files in TS to avoid adding code needs to be reworked in the future.

src/page/admin/auth/AutoCreate/AutoCreateForm.jsx
```diff
@@ -0,0 +1,96 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { FormControl, Grid, MenuItem, Switch, TextField, Select, InputLabel } from '@material-ui/core';
+
+function AutoCreateForm({ data = {}, setValue }) {
```
FredrikOseberg, line 5: Consider keeping the same naming convention on containing folder.

FredrikOseberg (approved):
LGTM. Added two small comments for consideration.

src/page/admin/auth/AutoCreate/AutoCreateForm.jsx
ivarconr: Agreed. Would '.tsx' ending be preferred?

src/page/admin/auth/AutoCreate/AutoCreateForm.jsx
ivarconr: would it then make sense to call the component "index.tsx" or would it be better with "AutoCreateForm/AutoCreateForm.tsx"?

src/page/admin/auth/AutoCreate/AutoCreateForm.jsx
FredrikOseberg: > Agreed. Would '.tsx' ending be preferred?

Yes, that is the convention for the rest of the typescript components we have in this project 👍

src/page/admin/auth/AutoCreate/AutoCreateForm.jsx
FredrikOseberg: AutoCreateForm.tsx would be preferred because it is much easier for the VS code search to locate that component when doing text search.

src/page/admin/auth/AutoCreate/AutoCreateForm.jsx
ivarconr: jikes. the event types required me to look up some docs 🤣

src/page/admin/auth/AutoCreate/AutoCreateForm.jsx
ivarconr: fixed

PR #327, opened 2021-08-23
Feat/feature routes

FredrikOseberg:


src/component/application/__tests__/__snapshots__/application-edit-component-test.js.snap
```diff
@@ -324,7 +324,7 @@ exports[`renders correctly with permissions 1`] = `
                       className="MuiTypography-root MuiListItemText-primary MuiTypography-body1 MuiTypography-displayBlock"
                     >
                       <a
-                        href="/features/strategies/ToggleA"
+                        href="/projects/undefined/toggles/ToggleA/strategies/ToggleA"
```
ivarconr, line 327: huh. "undefined"?

src/component/application/application-view.jsx
```diff
@@ -89,18 +89,21 @@ function ApplicationView({
...
                                 show={notFoundListItem({
-                                    createUrl: '/features/create',
+                                    createUrl: `/projects/${project}/toggles`,
                                     name,
                                     permission: CREATE_FEATURE,
                                     i,
                                 })}
                                 elseShow={foundListItem({
-                                    viewUrl: '/features/strategies',
+                                    viewUrl: `/projects/${project}/toggles/${name}/strategies`,
```
ivarconr, line 106: we must agree on `toggles` vs `features`. Currently the API will use `features`. I find it weird if we use other urls inside the UI. 

I would prefer the ability to prefix the url wiht "/api/admin" and get the JSON representation.

src/component/application/__tests__/__snapshots__/application-edit-component-test.js.snap
FredrikOseberg: Thanks. Feature toggle data in tests were missing "project" property.

src/component/application/application-view.jsx
FredrikOseberg: Redone to features.

ivarconr (approved):

PR #334, opened 2021-09-01
fix: sync

FredrikOseberg:


src/component/feature/FeatureToggleList/index.jsx
```diff
@@ -32,12 +32,19 @@ export const mapStateToPropsConfigurable = isFeature => state => {
...
         );
     }
-
     if (settings.filter) {
         try {
             const regex = new RegExp(settings.filter, 'i');
-            features = features.filter(
-                feature =>
+            features = features.filter(feature => {
+                if (!isFeature) {
```
ivarconr, line 39: huh, if not is feature?

FredrikOseberg: The isFeature property in the mapConfigurableProps determines whether or not it's we should treat it as a feature or archive. If it's not a feature, then it is an archived toggle and we should not check for strategies.

PR #336, opened 2021-09-17
feat: Add project and environment scoping to API keys

ivarconr:
Some help on improving the secrets column would be appreciated @FredrikOseberg


src/component/api-token/ApiTokenCreate/ApiTokenCreate.tsx
```diff
@@ -0,0 +1,184 @@
...
+        const value = event.target.value;
+        setData({...data, project: value})
+    }
+
+    const setEnvironment = (event: React.ChangeEvent<{value: string }>) => {
+        const value = event.target.value;
+        setData({...data, environment: value})
+    }
+
+    const selectableProjects = [{id: '*', name: 'ALL'}, ...(projects || [])].map(i => ({
```
FredrikOseberg, line 109: Are these necessary? Both hooks return an empty array if there is no data.

src/hooks/api/getters/useApiTokens/useApiTokens.ts
```diff
@@ -0,0 +1,33 @@
...
+const useApiTokens = () => {
+    const fetcher = async () => {
+        const path = formatApiPath(`api/admin/api-tokens`);
+        const res = await fetch(path, {
+            method: 'GET',
+        });
+        return res.json();
+    };
+
+    const { data, error } = useSWR(`api/admin/api-tokens`, fetcher);
```
FredrikOseberg, line 14: Just a minor detail, but extracting the key into a API_TOKENS_CACHE_KEY constant makes it less error prone, and allows us to export the key.

FredrikOseberg (approved):
Looks excellent. I only have minor comments.

src/component/api-token/ApiTokenCreate/ApiTokenCreate.tsx
ivarconr: A of course, did not think about what the hook actually returned. 

They will also slightly change now that you cannot select  "Any"  environment.

src/hooks/api/getters/useApiTokens/useApiTokens.ts
ivarconr: thanks 👍🏼

src/component/api-token/ApiTokenList/ApiTokenList.tsx
```diff
@@ -0,0 +1,208 @@
...
+        }
+        setDeleteToken(undefined);
+        setShowDelete(false);
+        refetch();
+        setToastData({
+            type: 'success',
+            show: true,
+            text: 'Successfully deleted API token.',
+        });
+        setTimeout(hideToast, 2000);
```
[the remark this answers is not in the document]
ivarconr: I was not aware!

PR #365, opened 2021-09-29
feat: project environments configuration

ivarconr:


src/component/project/ProjectEnvironment/ProjectEnvironment.tsx
```diff
@@ -0,0 +1,141 @@
...
+                onClick={refetch}
+                className={styles.apiError}
+                text="Error fetching environments"
+            />
+        );
+    };
+
+    const toggleEnv = async (env: ProjectEnvironment) => {
+        if(env.enabled) {
+            // await removeEnvironmentFromProject(id, env.name);
```
FredrikOseberg, line 56: Should this be run, or removed?

src/page/project/environment.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import ProjectEnvironment from '../../component/project/ProjectEnvironment/ProjectEnvironment';
+import PropTypes from 'prop-types';
+
+const render = ({ history }) => <ProjectEnvironment history={history} />;
```
FredrikOseberg, line 5: Is this needed? We could use useHistory inside of the component to access history and avoid the page abstraction.

src/component/project/ProjectEnvironment/ProjectEnvironment.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+        } else {
+            await addEnvironmentToProject(id, env.name);
+        }
+        refetch();
+    }
+
+    const handleDisableEnvironment = async () => {
+        if(selectedEnv && confirmName===selectedEnv.name) {
+            await removeEnvironmentFromProject(id, selectedEnv.name);
```
FredrikOseberg, line 67: Same here, should be wrapped in a try catch. Alternatively you could set propagateErrors to false and let the error be handled by the useApi parent hook. This will handle the most common usecases and set the error in the errors object exposed from the hook.

src/component/project/ProjectEnvironment/ProjectEnvironment.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { FormControlLabel, FormGroup, Switch } from '@material-ui/core';
+import useProjectApi from '../../../hooks/api/actions/useProjectApi/useProjectApi';
+import EnvironmentDisableConfirm from './EnvironmentDisableConfirm/EnvironmentDisableConfirm';
+
+export interface ProjectEnvironment {
+    name: string;
+    enabled: boolean;
+}
+
+const ProjectListNew = () => {
```
FredrikOseberg, line 26: Perhaps we could calls this ProjectEnvironmentList? We already have a ProjectList that serves a different purpose.

FredrikOseberg (changes requested):
Almost good to go! Just some small notes we might want to fix.

src/component/project/ProjectEnvironment/ProjectEnvironment.tsx
ivarconr: lol- removed!

src/component/project/ProjectEnvironment/ProjectEnvironment.tsx
```diff
@@ -0,0 +1,141 @@
...
+        );
+    };
+
+    const toggleEnv = async (env: ProjectEnvironment) => {
+        if(env.enabled) {
+            // await removeEnvironmentFromProject(id, env.name);
+            setSelectedEnv(env);
+
+        } else {
+            await addEnvironmentToProject(id, env.name);
```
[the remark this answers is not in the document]
ivarconr: woopsie. must fix! 👯

src/component/project/ProjectEnvironment/ProjectEnvironment.tsx
ivarconr: thanks for explaining this!

FredrikOseberg: No worries! I added propagation later because sometimes you want to access the error message if you are setting a toast or something similar.

src/page/project/environment.js
ivarconr: not needed

src/component/project/ProjectEnvironment/ProjectEnvironment.tsx
FredrikOseberg: in useApi.ts you can see the defaults: 

It checks for bad request, not found, unauthorized and forbidden. 

The keys are set as "unauthorized", "notFound", "badRequest", "forbidden". Now that I think about it, it is a bit convoluted that you have to check for the keys manually, but one could argue that you would like to perform different UI actions based on error you set. We could implement a general component that would resolve it for you. 

Lately, I have found the propagation easier, because in the component I generally know what feedback I would like to show.

ivarconr: I will go for the propagation for now.

PR #398, opened 2021-10-07
add info box in project page + make api ui responsive

ykhedher:
Info box:
image

API Keys (desktop):
image

API keys (mobile):
image


ivarconr:
Need to fix the e2e test before merging. 

Creating an API token should not be part of the feature spec. 

https://github.com/Unleash/unleash-frontend/pull/398/checks?check_run_id=3841209340

PR #410, opened 2021-10-11
fix: show project link in mobile nav

ykhedher:
- Changed the position of the project routes to the top so the project link appears the first one.
- Removed the flag from /projects.
Screen Shot 2021-10-11 at 9 59 08 AM


ykhedher:
Done ✅

PR #418, opened 2021-10-11
fix: show feature toggle type in list

ivarconr:
We should show the feature toggle type in all list of toggles.

Also cleaned up mobile view.


ivarconr:
[Add type icon to features list.](https://trello.com/c/OFIWomR0/483-add-type-icon-to-features-list)

PR #442, opened 2021-10-18
Fix/login redirect

FredrikOseberg:


FredrikOseberg:
[Bug: We do not redirect to login if session is lost. Also no indication to the user (you just get a empty list of everything).](https://trello.com/c/5VCnkqvj/516-bug-we-do-not-redirect-to-login-if-session-is-lost-also-no-indication-to-the-user-you-just-get-a-empty-list-of-everything)

PR #444, opened 2021-10-19
Mutate cache key for project when archiving feature.

chriswk:
- This causes a refetch of the feature list for project, preventing the
  just archived feature being present in the feature list when
  redirected back to the project feature list.


FredrikOseberg:
[Bug: We do not redirect to login if session is lost. Also no indication to the user (you just get a empty list of everything).](https://trello.com/c/5VCnkqvj/516-bug-we-do-not-redirect-to-login-if-session-is-lost-also-no-indication-to-the-user-you-just-get-a-empty-list-of-everything)

chriswk:
Thanks. I'll do that

PR #454, opened 2021-10-21
using copy library for copying user token

ykhedher:


ykhedher:
[Fix: Make GUI able to show API token in case copy does not work (ie: self-hosted / ff strict mode)](https://trello.com/c/eov5qRns/531-fix-make-gui-able-to-show-api-token-in-case-copy-does-not-work-ie-self-hosted-ff-strict-mode)

PR #479, opened 2021-10-29
Fix/strategy sidepanel

ykhedher:
Screen Shot 2021-11-01 at 2 09 07 PM


src/component/feature/FeatureView2/FeatureStrategies/FeatureStrategiesList/FeatureStrategyCard/FeatureStrategyCard.styles.ts
```diff
@@ -43,6 +43,12 @@ export const useStyles = makeStyles(theme => ({
     icon: {
         fill: theme.palette.primary.main,
     },
+    disabledButton: {
+        color: '#585858 !important',
```
FredrikOseberg, line 47: Why important?

ykhedher: Yes indeed it is important.

FredrikOseberg: There are usually better ways to do this than using `!important!`, because it can lead to unexpected behavior.

ykhedher: I am not sure, but I think if I use `.Mui-disabled { background-color: blue; }` . it will override all the styles for disabled button. So using important is like an exception for this case. If I remove the important nothing happens.

FredrikOseberg: I think you just need to override the disabled button styles: https://codesandbox.io/s/l9pk2wl69q

What is this supposed to be doing by the way? Because I believe that if we "fix" disabled style for one kind of IconButton we should do it for all of them.

ykhedher: I see. I think now it should be good. it is for the icons disabled in the side panel for strategies.

ykhedher:
[Strategy screen](https://trello.com/c/en7bospf/561-strategy-screen)

src/component/feature/FeatureView2/FeatureStrategies/FeatureStrategiesList/FeatureStrategyCard/FeatureStrategyCard.styles.ts
FredrikOseberg: Sorry for the miscommunication. These should not have any shadow after discussing with Elise we've elected to keep them as they were.

ykhedher:
[Strategy sidepanel](https://trello.com/c/gGZhgqtZ/559-strategy-sidepanel)

ykhedher:
[Strategy sidepanel](https://trello.com/c/gGZhgqtZ/559-strategy-sidepanel)

FredrikOseberg (approved):
LGTM. I think the last thing we could consider is to add a more visible outline to the sidebar button (the one with the conical gradient). Since we are reducing it's visual surface it is barely visible when the button is tabbed to.

PR #494, opened 2021-11-05
Fix/revive archived feature

ykhedher:
https://www.loom.com/share/14d256db30144b94a7376c4af6b17e7b


ykhedher:
[Should not be allowed to revive a toggle if project is deleted. [At least choose new project]](https://trello.com/c/2LBe9f7l/555-should-not-be-allowed-to-revive-a-toggle-if-project-is-deleted-at-least-choose-new-project)

PR #495, opened 2021-11-05
Fix/revive archived feature

ykhedher:
https://www.loom.com/share/14d256db30144b94a7376c4af6b17e7b

### DIFF
--- a/src/component/feature/FeatureToggleList/FeatureToggleListItem/FeatureToggleListItem.jsx
+++ b/src/component/feature/FeatureToggleList/FeatureToggleListItem/FeatureToggleListItem.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
 import { Link } from 'react-router-dom';
-import { Chip, IconButton, ListItem, Tooltip } from '@material-ui/core';
+import { Chip, ListItem, Tooltip } from '@material-ui/core';
 import { Undo } from '@material-ui/icons';
 
 import TimeAgo from 'react-timeago';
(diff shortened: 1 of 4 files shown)


ykhedher:
[Should not be allowed to revive a toggle if project is deleted. [At least choose new project]](https://trello.com/c/2LBe9f7l/555-should-not-be-allowed-to-revive-a-toggle-if-project-is-deleted-at-least-choose-new-project)

ykhedher:
[Should not be allowed to revive a toggle if project is deleted. [At least choose new project]](https://trello.com/c/2LBe9f7l/555-should-not-be-allowed-to-revive-a-toggle-if-project-is-deleted-at-least-choose-new-project)

src/component/feature/FeatureToggleList/FeatureToggleListItem/FeatureToggleListItem.jsx
```diff
@@ -31,11 +32,25 @@ const FeatureToggleListItem = ({
 }) => {
     const styles = useStyles();
 
+    const { projects } = useProjects();
     const isArchive = !!revive;
 
     const { name, description, type, stale, createdAt, project, lastSeenAt } =
         feature;
 
+    const isProjectDeleted = () => {
```
FredrikOseberg, line 41: Perhaps rename this to `projectExists`? The way it reads right now is a bit confusing because you are doing a conditional in reviveFeature based on the result of calling isProjectDeleted, usually you would expect a function worded like this to return true for the condition it's checking. 

If I'm reading reviveFeature now it looks like if the proejct is deleted the function will revive the feature, which is the opposite of what is happening.

src/component/feature/FeatureToggleList/FeatureToggleListItem/FeatureToggleListItem.jsx
```diff
@@ -126,7 +144,10 @@ const FeatureToggleListItem = ({
                     <ConditionallyRender
                         condition={hasAccess(UPDATE_FEATURE, project)}
                         show={
-                            <IconButton onClick={() => revive(feature.name)}>
+                            <IconButton
```
FredrikOseberg, line 147: We should use PermissionIconButton here, this action should only be allowed if you have access to UPDATE_FEATURE.

FredrikOseberg (changes requested):
Some small notes before we merge. Otherwise it looks goode.

FredrikOseberg (approved):
LGTM

PR #511, opened 2021-11-20
feat: add json payload validation in variants

ykhedher:
https://www.loom.com/share/78c30cdbb0184e5f9c5e5874ee41c610


src/component/feature/FeatureView2/FeatureVariants/FeatureVariantsList/AddFeatureVariant/AddFeatureVariant.tsx
```diff
@@ -204,6 +208,17 @@ const AddVariant = ({
         ]);
     };
 
+    const isJSON = jsonString => {
+        try {
+            let parsedString = JSON.parse(jsonString);
+            if (parsedString && typeof parsedString === 'object') {
```
FredrikOseberg, line 214: Unsure whether we'll allow it, but this is an example from MDN on JSON.parse: 

```
JSON.parse('"foo"')
```

It is technically valid JSON and is typeof string.

src/component/feature/FeatureView2/FeatureVariants/FeatureVariantsList/AddFeatureVariant/AddFeatureVariant.tsx
```diff
@@ -204,6 +208,17 @@ const AddVariant = ({
...
+                return true;
+            }
+        } catch (e) {}
```
FredrikOseberg, line 217: I think we should handle this more gracefully. Perhaps replace the catch with a finally clause and return false?

```
try { 
  ... 
} finally {
   return false
}
```

FredrikOseberg (changes requested):

src/component/feature/FeatureView2/FeatureVariants/FeatureVariantsList/AddFeatureVariant/AddFeatureVariant.tsx
FredrikOseberg: I think we want to allow any valid JSON on further reflection. Then the check here is uneccesary because JSON.parse will throw if it is not valid.

PR #517, opened 2021-11-24
fix: clear search input on route change

ykhedher:


src/component/feature/FeatureToggleList/FeatureToggleList.jsx
```diff
@@ -125,7 +129,6 @@ const FeatureToggleList = ({
         <div className={styles.featureContainer}>
             <div className={styles.searchBarContainer}>
                 <SearchField
-                    value={settings.filter}
```
FredrikOseberg, line 128: I think we would still want value to come from settings.filter?

FredrikOseberg (changes requested):

FredrikOseberg (approved):
LGTM

PR #519, opened 2021-11-24
fix/strategy tab

ykhedher:


FredrikOseberg (changes requested):
We need to remove the surrounding border, align the header text, and revert the separators for now until we can find a pattern that also works with constraints.

FredrikOseberg (approved):
LGTM!

PR #522, opened 2021-11-25
task: Add buttons for deleting/editing a constraint

sighphyre:
Adds two buttons onto the constraints listed in the strategies edit pane. The edit button will force an edit popup for all constraints, the delete will remove the constraint from browser memory and mark the constraints as dirty


src/component/common/Constraint/Constraint.tsx
```diff
@@ -1,31 +1,77 @@
...
+
+    const editable = deleteCallback && editCallback;
+
     return (
-        <div className={classes + ' ' + className} {...rest}>
-            <StringTruncator text={constraint.contextName} maxWidth="125" />
-            <FeatureStrategiesSeparator
-                text={constraint.operator}
-                maxWidth="none"
+        <div className={containerClasses + ' ' + className} {...rest}>
```
FredrikOseberg, line 44: you can import classnames and use that to join classes a little more smoothly: 

```
import classnames from 'classnames';

classnames(containerClasses, className)
```

FredrikOseberg (commented):
This looks good so far Simon, will wait for the updates we discussed earlier before making any further comments :)

src/component/common/Constraint/Constraint.tsx
sighphyre: That's super neat, thanks for the tip! Oddly enough, I had to restructure the CSS so I no longer need it but I'm going to file that in the back of my mind for next time!

FredrikOseberg (approved):
LGTM

PR #524, opened 2021-11-25
Autofocus dialog form fields, allow form submissions via pressing enter inside the form

thomasheartman:
This PR does two things: adds autofocus to forms contained within dialogs, and makes the same forms submittable when pressing `return`/`enter` on a keyboard.

It addresses [this trello issue](https://trello.com/c/JtCpopRp/602-disable-environment-modal-should-autofocus-input-field-submit-form-on-pressing-enter).

## solution / changes

For autofocus, it's just adding the `autofocus` attribute to the first input element in all dialog forms.

For submission via the keyboard, I have implemented that via the HTML5 `button` attribute `form`. This attribute lets you treat any button (even outside the form) as a submit button for the relevant form. You tie the button and the form together via the form's `id` attribute. To accomplish this, I have added an optional `formId` prop to the dialog component.

This prop triggers a few changes. When the prop is not falsy
- the component wraps the onClick handler in a function that first calls `preventDefault`. This is to stop page refreshes.
- the dialog's primary button is not given the `autofocus` attribute. This caused issues in some forms (such as the Create API Token form), where the button steals the focus that should be given to the first input field
- the primary button's type is set to `submit` instead of `button`. This is required for the form to use the button as a submit element. Without it, the form ignores the button and triggers a page refresh.

I have added ids to existing forms and passed them onto to the dialog.

This should not change anything for existing modals not containing forms or that have not been touched.

## discussion points

~~I have not added autofocus to forms where the first element is not a text input element. This should probably be fixed, but I wasn't sure whether it's something we wanted or not. After closer consideration, I think it's worth doing, but I'm happy to hear your thoughts on this.~~

*Update*: I _have_ added autofocus to all forms that I could find, though I may have missed something.

I have not updated ~~nested forms such as `AddUserForm`, `UserForm`, or~~ (see the update 👇🏾 ) the form for editing constraints. This is probably not a bad idea, but may require more thought, especially in relation to the autofocus attribute, which may not be desirable if the form is not presented within a modal.

I have not removed existing submit handlers, so there are some places where we both set `onSubmit={submit}` on the form _and_ `onClick={submit}` on the dialog component. In these cases, we should be able to safely remove the form's submit handler, but I'd like to hear whether you agree before I do that.

## questions

Do we have a storybook or something similar where I can see _all_ components affected by this? I can test the dialogs I'm aware of, but it's hard to find absolutely everything.

-----

Anything else I should comment on or have forgotten; let me know! Any feedback is very welcome 😄

-----

## Edit / update

Since I opened this PR, I've also added a few more autofocus points: 
- Adding or updating users now gives autofocus.
- The form for creating new toggles also steals focus to the name field


thomasheartman:
[Disable environment modal should autofocus input field, submit form on pressing `enter`](https://trello.com/c/JtCpopRp/602-disable-environment-modal-should-autofocus-input-field-submit-form-on-pressing-enter)

chriswk (approved):
Nice job :+1:

thomasheartman:
Update: I've added autofocus to a few more places than the OP mentions. I've now added it for the dialogs:
- add new user
- update user

and for the new create feature page (`FeatureCreate.tsx`). 

Of these, the last is probably the most controversial, and I'm happy to remove it if you think that's better. However, here's my reasoning:
- That form is only used in one place (according to my language server at least)
- There's nothing else on that page (save for headers and footers)
- Because of the above, if the user is there, it's probably because they really want to create a toggle
- Automatically moving the focus there makes it quicker and easier (especially as we don't have skip links ('skip to main content') to make navigating there easier)
- We also don't have a `main` section on the page, so landmark navigation doesn't take you there quickly.

I'll update the OP too.

src/component/common/Dialogue/Dialogue.tsx
```diff
@@ -61,12 +69,14 @@ const Dialogue: React.FC<IDialogue> = ({
...
+                            form={formId}
                             color="primary"
                             variant="contained"
-                            onClick={onClick}
-                            autoFocus
+                            onClick={handleClick}
+                            autoFocus={!formId}
                             disabled={disabledPrimaryButton}
                             data-test={DIALOGUE_CONFIRM_ID}
+                            type={formId ? 'submit' : undefined}
```
FredrikOseberg, line 79: Why undefined here instead of button? I think material ui will resolve button as type if it is undefined, but if it does we might as well skip the implicit overhead and be explicit?

FredrikOseberg (approved):
Quite elegant solution. Shouldn't impact anything else.

src/component/common/Dialogue/Dialogue.tsx
thomasheartman: That's a good point. I did initially set it to be 'button' instead of undefined because that's what Material does by default. However, I decided to change it to `undefined` in the event that Material changes its defaults later. Maybe they'd like to use a different type instead. If we pass `undefined`, then we let Material choose for us.

However, I also don't see this happening any time soon (why would it?), so I'm happy to change it back to 'button' if you think that's better 💁🏾 . 

One final point, though: I reckon most uses of Material buttons won't define the type at all, so it'll be 'undefined' in a large majority of cases. They have probably optimized well for that use case, so I don't expect it to have much of an impact (but I haven't measured it!).

FredrikOseberg: I'll merge this now, so I just made the change. Those are fair points, I think it is entirely fine to pass undefined and let material set the button type in most use cases, however when we elect to explicitly set the type, it seems counter-intuitive to me to pass undefined as a value when the spec is explicit in terms of accepted values. 

In the event that material changes how they handle internal type settings, I don't think they will ever elect to override a specific choice that complies with the HTML spec.

When it comes to performance, it depends on how they have implemented it. But it's still such a small optimization that I don't think it's even worth measuring.

thomasheartman: Sure; I can get on board with that! 😄 Just merge at your convenience, then 🚂

PR #525, opened 2021-11-25
Fix variant list

sighphyre:
- Fix an issue where clicking a variant in the list would popup a modal
dialog with empty details


ivarconr (approved):
Easy!

FredrikOseberg (approved):
LGTM. But as the tests are now breaking, maybe the easiest course of action would be to remove this onClick? I think using the edit button would be sufficient. It's not really accessible either since it's added to a table element.

PR #526, opened 2021-11-26
docs: Add some details on running and debugging the e2e tests

sighphyre:


thomasheartman (approved):
Super! 🎉  Nice and clear; looks easy to follow! Only one comment, which you can choose whether to address or not ☺️

FredrikOseberg (changes requested):
Nice! Lovely to have this, one thing I'd like to have in this PR is how to run it agains the heroku instance (detailing the nuance with the different password authentication), because that is the only way you can replicate the environment of the CI server.

FredrikOseberg (approved):
LGTM

PR #566, opened 2021-12-20
fix: adjust main splash container when zoom in

ykhedher:


ykhedher:
[The environments modal breaks when user zooms in](https://trello.com/c/RohMTvgk/688-the-environments-modal-breaks-when-user-zooms-in)

FredrikOseberg (approved):
LGTM

PR #576, opened 2021-12-30
feat: add show password for all passwords input

ykhedher:
I know that this task was not in the backlog, but I really needed this feature when I log in with a new user.

Screen Shot 2021-12-30 at 10 59 16 AM


ykhedher:
[Add show password for all passwords input](https://trello.com/c/D1bvEDtY/690-add-show-password-for-all-passwords-input)

src/component/admin/users/change-password-component.jsx
```diff
@@ -20,8 +27,17 @@ function ChangePassword({
...
     const [error, setError] = useState({});
     const [validPassword, setValidPassword] = useState(false);
+    const [showPassword, setShowPassword] = useState(false);
     const commonStyles = useCommonStyles();
 
+    const handleClickShowPassword = () => {
+        setShowPassword(!showPassword);
+    };
+
+    const handleMouseDownPassword = e => {
```
FredrikOseberg, line 37: What is the function of this?

ykhedher: Just following the Material UI example here: https://mui.com/components/text-fields/#input-adornments

FredrikOseberg (changes requested):
I would prefer this to be encapsulated in a single PasswordField component that handles the usecase instead of repeating the functionality four times. It gives us a single point of entry for change later on.

ykhedher:
> I would prefer this to be encapsulated in a single PasswordField component that handles the usecase instead of repeating the functionality four times. It gives us a single point of entry for change later on.

Totally agree 👍

FredrikOseberg (approved):
Cool. Looks good to me!

PR #601, opened 2022-01-17
feat: add edit and create user screen (NEW)

ykhedher:


ykhedher:
[Create screen - User](https://trello.com/c/UIU2K0Wo/683-create-screen-user)

FredrikOseberg (approved):
LGTM once we change the naming of the component and update the docs links.

ykhedher:
> Hm. I see we have now lost the ability to copy the token. We need to have this in case email configuration is not set up. I think we should have that as a confirm screen after we create the user.

I forgot about that. I was going to ask about it, but I think I forgot since it is 15 days ago lol

FredrikOseberg (approved):
LGTM

PR #602, opened 2022-01-17
feat: add create and edit project screen (NEW)

ykhedher:


ykhedher:
[Create screen - Project](https://trello.com/c/4LK8V8vH/684-create-screen-project)

src/component/project/Project/hooks/useProjectForm.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    const getProjectPayload = () => {
+        return {
+            id: projectId,
+            name: projectName,
+            description: projectDesc,
+        };
+    };
+
+    const validateIdUniqueness = () => {
```
FredrikOseberg, line 40: This is not good enough to validate the id. We need to call an endpoint to do this on the server side. The existing project performs the validation like this: 

```
    const validateId = async (id: string) => {
        if (editMode) return true;

        const e = { ...errors };
        try {
            await props.validateId(id);
            e.id = undefined;
        } catch (err: any) {
            e.id = err.message;
        }

        setErrors(e);
        if (e.id) return false;
        return true;
    };

```

src/component/project/Project/ProjectForm/ProjectForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+                />
+            </div>
+
+            <div className={styles.buttonContainer}>
+                <Button onClick={handleCancel} className={styles.cancelButton}>
+                    Cancel
+                </Button>
+                <PermissionButton
+                    onClick={handleSubmit}
+                    permission={ADMIN}
```
FredrikOseberg, line 92: Only admins can create project? This should be CREATE_PROJECT permission.

src/component/project/Project/ProjectForm/ProjectForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    type="submit"
+                >
+                    {submitButtonText} project
```
FredrikOseberg, line 95: This should no longer be needed. We have separate components now for Create and Edit.

src/component/project/Project/ProjectForm/ProjectForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+    return (
+        <form onSubmit={handleSubmit} className={styles.form}>
+            <h3 className={styles.formHeader}>Project Information</h3>
+
+            <div className={styles.container}>
+                <p className={styles.inputDescription}>
+                    What is your project Id?
+                </p>
+                <Input
+                    className={styles.input}
```
FredrikOseberg, line 48: We need to make sure this field does not allow spaces. Currently we are doing it like this in the old view: 

```
    <TextField
          label="Project Id"
          name="id"
          placeholder="A-unique-key"
          value={project.id}
          error={!!errors.id}
          helperText={errors.id}
          disabled={editMode}
          variant="outlined"
          size="small"
          onBlur={v => validateId(v.target.value)}
          onChange={v =>
              setValue('id', trim(v.target.value))
          }
      />
```

FredrikOseberg (changes requested):
Needs a few changes before it's ready.

PR #618, opened 2022-01-19
fix: UI crashes when changing parameter name when the param is of type number

ykhedher:


ykhedher:
[The admin UI crashes when you rename a number-type parameter for a custom strategy](https://trello.com/c/SaG3peL8/758-the-admin-ui-crashes-when-you-rename-a-number-type-parameter-for-a-custom-strategy)

FredrikOseberg:
This has been fixed in another PR.

PR #621, opened 2022-01-21
fix: project access

ykhedher:


src/component/project/ProjectAccess/ProjectAccess.tsx
```diff
@@ -18,54 +18,41 @@ import {
...
-        } catch (e) {
-            setToastApiError(e.toString());
-        }
-    };
-
     if (isOss()) {
         return (
-            <PageContent>
+            <PageContent
+                headerContent={<HeaderTitle title="Projecttt Access" />}
```
FredrikOseberg, line 54: Projecttt accesss?

src/component/project/ProjectAccess/ProjectAccess.tsx
```diff
@@ -78,58 +63,36 @@ const ProjectAccess = () => {
         );
     }
 
-    const handleRoleChange = (userId, currRoleId) => async evt => {
-        const roleId = evt.target.value;
-        try {
-            await projectApi.removeUserFromRole(id, currRoleId, userId);
-            await projectApi.addUserToRole(id, roleId, userId).then(() => {
+    const handleRoleChange =
+        (userId: string, currRoleId: string) => async evt => {
```
ivarconr, line 67: Would be good to also have the event (`evt`) typed.

src/component/project/ProjectAccess/ProjectAccess.tsx
```diff
@@ -78,58 +63,35 @@ const ProjectAccess = () => {
...
-                type: 'error',
-                title: err.message || 'Server problems when adding users.',
-            });
-        }
-    };
+            }
+        };
 
-    const removeAccess = (userId: number, roleId: number) => async () => {
+    const removeAccess = (userId: string, roleId: string) => async () => {
```
ivarconr, line 86: I generally find it a bit strange that useId and roleId are strings when I "know" they are numbers inside unleash. But it should not matter what they are from the frontend perspective, so maybe strings are better.

src/component/project/ProjectAccess/ProjectAccess.tsx
```diff
@@ -78,58 +63,35 @@ const ProjectAccess = () => {
...
-            await projectApi.removeUserFromRole(id, roleId, userId).then(() => {
-                setToastData({
-                    type: 'success',
-                    title: 'User have been removed from project',
-                });
+            await removeUserFromRole(id, roleId, userId);
+            refetchProjectAccess();
+            setToastData({
+                type: 'success',
+                title: 'User have been removed from project',
```
ivarconr, line 92: Typo (also in pre):
```suggestion
                title: 'The user has been removed from project',
```

src/component/project/ProjectAccess/ProjectAccessAddUser/ProjectAccessAddUser.tsx
```diff
@@ -11,14 +9,35 @@ import {
...
+interface IProjectAccessAddUserProps {
+    roles: IRole[];
+}
+
+const ProjectAccessAddUser = ({ roles }: IProjectAccessAddUserProps) => {
+    const { id } = useParams<{ id: string }>();
     const [user, setUser] = useState();
-    const [role, setRole] = useState({});
+    const [role, setRole] = useState<IRole>({
+        id: 0,
```
ivarconr, line 28: defaulting to 0 can be really bad. This might even be the admin role...

ivarconr (approved):

src/component/project/ProjectAccess/ProjectAccess.tsx
FredrikOseberg: I agree. We should try to keep it as consistent as possible.

FredrikOseberg (changes requested):
Found some quite serious bugs with this PR that needs to be addressed before merging. 

(1) Changing a project members role will fail intermittently, which causes the member to be removed from the project.
(2) Adding a member by clicking the dropdown instead of using tab will not select that member and you will get an error when adding them (this will also spew out database information, which we should address in unleash-server). 

Also, it might be worth discussing whether or not unleash should restore the users previous role if for any reason the operation fails halfway.

ykhedher:
[Fix: Project access component](https://trello.com/c/hHnbQlcA/792-fix-project-access-component)

src/component/project/ProjectAccess/ProjectAccessAddUser/ProjectAccessAddUser.tsx
```diff
@@ -0,0 +1,241 @@
...
+                                onChange={handleQueryUpdate}
+                                InputProps={{
+                                    ...params.InputProps,
+                                    startAdornment: (
+                                        <InputAdornment position="start">
+                                            <Search />
+                                        </InputAdornment>
+                                    ),
+                                    endAdornment: (
+                                        <React.Fragment>
```
olav, line 199: `<>`.

src/component/project/ProjectAccess/ProjectAccessAddUser/ProjectAccessAddUser.tsx
```diff
@@ -0,0 +1,241 @@
...
+                        freeSolo
+                        getOptionSelected={() => true}
+                        filterOptions={o => o}
+                        getOptionLabel={(option: IProjectAccessUsers) => {
+                            if (option) {
+                                return `${option.name || '(Empty name)'} <${
+                                    option.email || option.username
+                                }>`;
+                            } else return '';
+                        }}
```
olav, line 180: Would put this fn in a `const` outside the `return`, for readability.

src/component/project/ProjectAccess/ProjectAccessAddUser/ProjectAccessAddUser.tsx
```diff
@@ -0,0 +1,241 @@
...
+    };
+
+    const handleSelectUser = (
+        evt: ChangeEvent<{}>,
+        selectedUser: string | IProjectAccessUsers | null
+    ) => {
+        setOptions([]);
+        if (typeof selectedUser === 'string' || selectedUser === null) {
+            return;
+        }
```
olav, line 96: Seems like the next `if` would cover these cases as well.

src/component/project/ProjectAccess/ProjectAccessAddUser/ProjectAccessAddUser.tsx
```diff
@@ -0,0 +1,241 @@
...
+        setLoading(false);
+    };
+
+    const handleQueryUpdate = (evt: { target: { value: string } }) => {
+        const q = evt.target.value;
+        search(q);
+        if (options.length === 1) {
+            return;
+        }
```
olav, line 79: Removable?

src/component/project/ProjectAccess/ProjectAccessAddUser/ProjectAccessAddUser.tsx
```diff
@@ -0,0 +1,241 @@
...
+            const result = await searchProjectUser(query);
+            const userSearchResults = await result.json();
+
+            const filteredUsers = userSearchResults.filter(
+                (selectedUser: IProjectAccessUsers) => {
+                    const selected = access.users.find(
+                        (user: IProjectAccessUsers) =>
+                            user.id === selectedUser.id
+                    );
+                    return selected ? false : true;
```
olav, line 64:
```
return !selected
```

olav (changes requested):

src/component/common/PaginateUI/PaginateUI.tsx
```diff
@@ -15,6 +15,7 @@ interface IPaginateUIProps {
     prevPage: () => void;
     setPageIndex: (idx: number) => void;
     nextPage: () => void;
+    style?: { [index: string]: any };
```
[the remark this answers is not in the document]
FredrikOseberg: Perfect.

src/component/common/PermissionIconButton/PermissionIconButton.tsx
```diff
@@ -2,17 +2,23 @@ import { IconButton, Tooltip } from '@material-ui/core';
...
     onClick?: (e: any) => void;
     projectId?: string;
     environmentId?: string;
+    edge?: string;
+    className?: string;
+    title?: string;
 }
 
-const PermissionIconButton: React.FC<IPermissionIconButtonProps> = ({
+const PermissionIconButton: React.FC<IProps> = ({
```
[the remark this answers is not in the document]
FredrikOseberg: You're right. I was playing with the idea of changing the convention to IProps, the main benefit of having the component name in the interface is that it's searchable - which is kind of redundant with the filename being so easily searchable. But if we do this we should do it across everything, so I'll revert this.

src/component/project/ProjectAccess/ProjectAccess.tsx
```diff
@@ -1,71 +1,49 @@
...
+import { ProjectAccessList } from './ProjectAccessList/ProjectAccessList';
 
-const ProjectAccess = () => {
-    const { id } = useParams<IFeatureViewParams>();
+export const ProjectAccess = () => {
+    const { id: projectId } = useParams<IProjectViewParams>();
     const styles = useStyles();
-    const [roles, setRoles] = useState([]);
-    const [users, setUsers] = useState([]);
     const [error, setError] = useState();
```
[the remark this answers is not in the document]
FredrikOseberg: Nice catch. This is totally correct and allows us to remove even more code.

src/component/common/PermissionIconButton/PermissionIconButton.tsx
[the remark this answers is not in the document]
olav: That makes sense. Either way works for me, but true, the consistency is nice.

src/component/project/ProjectAccess/ProjectAccess.tsx
```diff
@@ -78,58 +56,49 @@ const ProjectAccess = () => {
...
-    const handleRoleChange = (userId, currRoleId) => async evt => {
-        const roleId = evt.target.value;
-        try {
-            await projectApi.removeUserFromRole(id, currRoleId, userId);
-            await projectApi.addUserToRole(id, roleId, userId).then(() => {
+    const handleRoleChange =
+        (userId: number, currRoleId: number) =>
+        async (
+            evt: React.ChangeEvent<{
+                name?: string | undefined;
```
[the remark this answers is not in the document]
FredrikOseberg: Yep

src/component/project/ProjectAccess/ProjectAccessAddUser/ProjectAccessAddUser.tsx
FredrikOseberg: Agree

src/component/project/ProjectAccess/ProjectAccessAddUser/ProjectAccessAddUser.tsx
FredrikOseberg: Good point. I'll fix it.

src/component/project/ProjectAccess/ProjectAccessAddUser/ProjectAccessAddUser.tsx
FredrikOseberg: Normally. But because of a quirk on how the Material-UI <AutomComplete /> component works you can either these values passed in: `selectedUser: string | IProjectAccessUser | null`

If I remove those checks then ts will complain that "id" doesn't exist on string, and I can't force the AutoComplete field to conform to my handler (at least I haven't found a way).

src/component/project/ProjectAccess/ProjectAccessAddUser/ProjectAccessAddUser.tsx
FredrikOseberg: Yeah, I don't see the point of this either.

FredrikOseberg (approved):

PR #627, opened 2022-01-24
feat: update dialog to accept permissions

ykhedher:
https://www.loom.com/share/3903a92d376a4dbe87849a5369968b3c

### DIFF
--- a/src/component/common/Dialogue/Dialogue.tsx
+++ b/src/component/common/Dialogue/Dialogue.tsx
@@ -10,7 +10,7 @@ import {
 import ConditionallyRender from '../ConditionallyRender/ConditionallyRender';
 import { useStyles } from './Dialogue.styles';
 import { DIALOGUE_CONFIRM_ID } from '../../../testIds';
-import PermissionButton from '../PermissionButton/PermissionButton';
+import { IPermissionIconButtonProps } from '../PermissionButton/PermissionButton';
 
 interface IDialogue {
     primaryButtonText?: string;
@@ -24,7 +24,7 @@ interface IDialogue {
     maxWidth?: 'lg' | 'sm' | 'xs' | 'md' | 'xl';
     disabledPrimaryButton?: boolean;
     formId?: string;
-    permission?: string;
+    permissionButton?: React.ReactNode;
 }
 
 const Dialogue: React.FC<IDialogue> = ({
@@ -39,7 +39,7 @@ const Dialogue: React.FC<IDialogue> = ({
     maxWidth = 'sm',
     fullWidth = false,
     formId,
-    permission,
+    permissionButton,
 }) => {
     const styles = useStyles();
     const handleClick = formId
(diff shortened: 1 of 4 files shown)


ykhedher:
[RBAC Environments - Access to toggling a feature on in an environment triggers the popup that takes you to add a strategy even if you don't have CREATE_FEATURE_STRATEGY permission.](https://trello.com/c/i57MI5Nu/749-rbac-environments-access-to-toggling-a-feature-on-in-an-environment-triggers-the-popup-that-takes-you-to-add-a-strategy-even-if)

src/component/common/EnvironmentStrategiesDialog/EnvironmentStrategyDialog.tsx
```diff
@@ -29,6 +29,7 @@ const EnvironmentStrategyDialog = ({
             onClose={() => onClose()}
             title="You need to add a strategy to your toggle"
             primaryButtonText="Take me directly to add strategy"
+            permission={UPDATE_FEATURE}
```
FredrikOseberg, line 32: This is not the correct permission. The permission here is UPDATE_FEATURE_STRATEGY, and it needs to take into account the projectId and the environment for that permission.

src/component/common/Dialogue/Dialogue.tsx
```diff
@@ -68,18 +71,37 @@ const Dialogue: React.FC<IDialogue> = ({
...
-                            variant="contained"
-                            onClick={handleClick}
-                            autoFocus={!formId}
-                            disabled={disabledPrimaryButton}
-                            data-test={DIALOGUE_CONFIRM_ID}
-                            type={formId ? 'submit' : 'button'}
-                        >
-                            {primaryButtonText || "Yes, I'm sure"}
-                        </Button>
+                        <ConditionallyRender
```
FredrikOseberg, line 74: Perhaps we could simply pass an entire PermissionButton in instead of the permission? Otherwise we might have to accept a lot of arguments that are not specific to the dialogue, such as projectId and environment - which will be available in the enclosing scope of the component that is actually using the dialogue component.

src/component/common/Dialogue/Dialogue.tsx
```diff
@@ -68,18 +70,24 @@ const Dialogue: React.FC<IDialogue> = ({
...
+                            condition={Boolean(permissionButton)}
+                            show={permissionButton}
+                            elseShow={
```
FredrikOseberg, line 76: This button previously only showed if there was an onClick handler passed. With the current code it will always show, even if an onClick handler is not passed to the dialogue.

FredrikOseberg (changes requested):
Need to add back the condition for the button.

src/component/common/EnvironmentStrategiesDialog/EnvironmentStrategyDialog.tsx
```diff
@@ -29,15 +30,23 @@ const EnvironmentStrategyDialog = ({
             onClose={() => onClose()}
             title="You need to add a strategy to your toggle"
             primaryButtonText="Take me directly to add strategy"
+            permissionButton={
+                <PermissionButton
+                    permission={CREATE_FEATURE_STRATEGY}
+                    projectId={projectId}
```
FredrikOseberg, line 36: This is still lacking environment property

FredrikOseberg: This needs to be the environment that the user is currently trying to activate.

FredrikOseberg (approved):
Looks good now.

PR #629, opened 2022-01-25
feat: allow enterprise to disable password based login

ivarconr:


src/component/admin/auth/simple-auth.tsx
```diff
@@ -0,0 +1,96 @@
...
+                        <p>Allow users to login with username & password</p>
+                    </Grid>
+                    <Grid item md={6} style={{ padding: '20px' }}>
+                        <FormControlLabel
+                            control={
+                                <Switch
+                                    onChange={updateDisabled}
+                                    value={!disabled}
+                                    name="disabled"
+                                    checked={!disabled}
```
ivarconr, line 71: I know this one looks a bit strange, but it is on purpose. We want to store the value as "I have actively disabled password based login", and thus call the config option "disabled". 

But presenting it as disabled to the user can be a bit confusing, so we invert it to "Is password based authentication enabled".

ivarconr:
[feat: SSO customers should have the possibility to disable the ability to login with password (iteration 1)](https://trello.com/c/GoIK8zjC/276-feat-sso-customers-should-have-the-possibility-to-disable-the-ability-to-login-with-password-iteration-1)

src/component/admin/auth/simple-auth.tsx
```diff
@@ -0,0 +1,96 @@
...
+    Switch,
+} from '@material-ui/core';
+import { Alert } from '@material-ui/lab';
+import PageContent from '../../common/PageContent/PageContent';
+import AccessContext from '../../../contexts/AccessContext';
+import { ADMIN } from '../../providers/AccessProvider/permissions';
+import useAuthSettings from '../../../hooks/api/getters/useAuthSettings/useAuthSettings';
+import useAuthSettingsApi, {ISimpleAuthSettings } from '../../../hooks/api/actions/useAuthSettingsApi/useAuthSettingsApi';
+
+function SimpleAuth() {
```
ykhedher, line 15: It will be good if we conserve the same pattern of how we create our components. Which are all arrow function components.

ykhedher (approved):
Looks good to me. 👍

src/component/admin/auth/simple-auth.tsx
FredrikOseberg: Good point. I would also prefer if we kept the naming convention of components `simple-auth.tsx -> SimpleAuth.tsx`.

src/component/admin/auth/simple-auth.tsx
FredrikOseberg: Yeah. I prefer when I look at the component to understand what is going on immediately. Disabled has a strong meaning in the web development world, that it's difficult for me to associate it with something else. I just know I will look at this component at some point and wonder why we are setting disabled as the value. Perhaps we could call it `[disablePasswordAuth, setDisablePasswordAuth]` ?

src/component/admin/auth/simple-auth.tsx
```diff
@@ -0,0 +1,96 @@
...
+                <Grid container spacing={3}>
+                    <Grid item md={12}>
+                        <Button
+                            variant="contained"
+                            color="primary"
+                            type="submit"
+                        >
+                            Save
+                        </Button>{' '}
+                        <small>{info}</small>
```
FredrikOseberg, line 87: Why not toggle this loading based on the free loading state you get from your API hook instead of using an internal state?

src/component/admin/auth/simple-auth.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    
+    const onSubmit = async e => {
+        e.preventDefault();
+        
+        setInfo('...saving');
+        try {
+            const settings: ISimpleAuthSettings = { disabled };
+            await updateSettings(settings);
+            setInfo('Settings stored');
```
FredrikOseberg, line 48: Add a setToastData here using the useToast to triger a toast on successful update?

src/component/admin/auth/simple-auth.tsx
```diff
@@ -0,0 +1,96 @@
...
+            setTimeout(() => setInfo(''), 2000);
+        } catch (e) {
```
FredrikOseberg, line 50: You could set an error here with the setToastApiError.

FredrikOseberg (approved):
LGTM. Nothing major, but a few low hanging fruits that you could add to improve the experience and continue the path forward in terms of aligning conventions.

src/component/admin/auth/simple-auth.tsx
ivarconr: thanks, make sense!

src/component/admin/auth/simple-auth.tsx
ivarconr: `PasswordAuthSettings.tsx` is probably more accurate!

src/component/admin/auth/simple-auth.tsx
ivarconr: because I am not fluent in the useSwr hooks yet, and it was partially based on the existing code. Will take a stab at this!

src/component/admin/auth/simple-auth.tsx
ivarconr: done!

src/component/admin/auth/simple-auth.tsx
ivarconr: done.

PR #635, opened 2022-01-27
feat: create useAuthApi hook and use it in Authentication component

ykhedher:


olav:
New branch: https://github.com/Unleash/unleash-frontend/tree/refactor-login

PR #637, opened 2022-01-27
Refactor/users list

ykhedher:


ykhedher:
[refactor: usersList to hooks and add toast on user Deletion](https://trello.com/c/ADlZvkxC/804-refactor-userslist-to-hooks-and-add-toast-on-user-deletion)

PR #651, opened 2022-02-01
Fix/cleanup unused code

FredrikOseberg:
* Rename FeatureView2 to FeatureView
* Fix path to go to /project/:projectId/features/:featureId
* Update paths in application
* Add redirect on /features2 to /project/:projectId/features/:featureId
* Remove unused code
* Refactor RedirectFeatures to use hooks instead of redux


olav (commented):
LGTM AFAIK, but probably best if someone else takes a look.

src/hooks/api/getters/useFeatures/useFeatures.ts
```diff
@@ -0,0 +1,40 @@
+import useSWR, { mutate, SWRConfiguration } from 'swr';
+import { useState, useEffect } from 'react';
+import { formatApiPath } from '../../../../utils/format-path';
+import handleErrorResponses from '../httpErrorResponseHandler';
+
+const useFeatures = (options: SWRConfiguration = {}) => {
+    const fetcher = async () => {
+        const path = formatApiPath('api/admin/features/');
```
FredrikOseberg, line 8: Should move this out of the fetcher. We don't need to recreate the path on every fetch.

olav: True! I've moved the path out in my branch, so I can fix when doing the rebase:

https://github.com/Unleash/unleash-frontend/pull/663/files#diff-b4d115db9cc2bc0a0e4c3d8a5f74cdac2c83fa528ca19dde4bbdb1bffd328aaeR7

src/component/feature/RedirectFeatureView/RedirectFeatureView.tsx
```diff
@@ -1,30 +1,35 @@
...
-        }
         /* eslint-disable-next-line */
-    }, []);
+        const toggle = features.find(
+            (toggle: IFeatureToggle) => toggle.name === name
+        );
+
+        setFeatureToggle(toggle);
+        /* eslint-disable-next-line */
+    }, [features]);
```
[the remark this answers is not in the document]
FredrikOseberg: True, and we probably should though the name will either be in params or not and shouldn't change. I guess we'd also need to add the setState function in order to remove the annotation and wrap that in a function with useCallback.

PR #652, opened 2022-02-01
Fix/login redirect set password

FredrikOseberg:
SWRProvider is set to redirect on 401, however the reset-password and new-user routes are only accessible if you are not logged in, which causes a redirect away from these pages when they are necessary. This was introduced because we started fetching the user on an interval, which will continously trigger 401 if the user is not logged in and was part of the 4.6.0 release


src/component/providers/SWRProvider/SWRProvider.tsx
```diff
@@ -22,17 +22,23 @@ const SWRProvider: React.FC<ISWRProviderProps> = ({
...
+                path === '/login' ||
+                path === '/new-user' ||
+                path === '/reset-password'
+            ) {
                 return;
             }
 
+            cache.clear();
+
             history.push('/login');
```
sighphyre, line 42: I just got here and this code is new to me butt is this still correct even though the check has changed from just /login to checking against any of /login, /new-user and /reset-password?

sighphyre (approved):
I have a question but aside from that, looks good - nice catch!

ykhedher (approved):
LGTM

src/component/providers/SWRProvider/SWRProvider.tsx
FredrikOseberg: Yeah. So if you are on any of those pages, then we want to keep whatever is in the useSWR cache because the cache represents every getter hooks cache, and we use the cache to keep track of data fetched from the API like `useResetPassword`. If we are not on any of those pages, then we clear the entire cache of the application and redirect to login.

FredrikOseberg:
As a side note, this functionality is finicky and we don't often use it ourselves. I think this warrants an e2e test.

src/component/providers/SWRProvider/SWRProvider.tsx
sighphyre: Makes sense, thanks!

PR #653, opened 2022-02-01
refactor: port ApiDetails to useSWR and TS

olav:


src/component/menu/ApiDetails/ApiDetails.tsx
```diff
@@ -0,0 +1,106 @@
+import useUiConfig from '../../../hooks/api/getters/useUiConfig/useUiConfig';
+import { ReactElement } from 'react';
+import ConditionallyRender from '../../common/ConditionallyRender';
+import { IVersionInfo } from '../../../interfaces/uiConfig';
+
+interface PartialUiConfig {
```
FredrikOseberg, line 6: Previously in the frontend we've tried to follow a convention where we prepend each interface with an I so that we can easily recognize interfaces vs types. It's just a nice mental shortcut that helps distinguish an interface from a type in the code. `IPartialUiConfig` or something similar would follow that convention.

src/component/menu/ApiDetails/ApiDetails.tsx
```diff
@@ -0,0 +1,106 @@
...
+	name: string;
+	version: string;
+	slogan?: string;
+	environment?: string;
+	versionInfo?: IVersionInfo;
+}
+
+interface Props {
```
FredrikOseberg: I'm a bit back and forth on this, but we've used ComponentNameProps as a convention for props. I think we could easily just call it Props when it lives with a component, especially since it's not exported anywhere. But I would like us to follow the same convention.

src/component/menu/ApiDetails/ApiDetails.tsx
```diff
@@ -0,0 +1,106 @@
...
+	uiConfig: PartialUiConfig;
+}
+
+export function ApiDetailsContainer(): ReactElement {
```
FredrikOseberg, line 18: Nice. This makes this component super testable. But perhaps we could just use the hook directly in the Footer and pass in the data from there?

src/component/menu/ApiDetails/ApiDetails.tsx
```diff
@@ -0,0 +1,106 @@
...
+	uiConfig: IPartialUiConfig;
+}
+
+export function ApiDetailsContainer(): ReactElement {
+	const { uiConfig } = useUiConfig();
+
+	return <ApiDetails uiConfig={uiConfig} />;
+}
+
+export function ApiDetails(props: IProps): ReactElement {
```
FredrikOseberg, line 24: We're trying to follow a convention of using the arrow function pattern when it comes to components, just to keep it consistent across the repository: 

```
const ApiDetails = () => {
    ...
}
```

src/component/menu/ApiDetails/ApiDetails.tsx
```diff
@@ -0,0 +1,106 @@
...
+			<UpdateNotification {...props} />
+			<br />
+			<small>{props.uiConfig.slogan}</small>
+			<br />
+			<InstanceId {...props} />
+		</section>
+	);
+}
+
+function InstanceId(props: IProps) {
```
FredrikOseberg, line 37: I think this is fine in this case, but personally I prefer when a component has it's own file. I think it's either a case of (a) a component is large enough to warrant splitting it out into multiple components, which will then be placed in it's own file or (b) the amount of code is to small to make a meaningful abstraction. In this case I think the code is so small that we could just merge it directly into ApiDetails without the component abstractions. What do you think?

src/component/menu/ApiDetails/ApiDetails.tsx
```diff
@@ -0,0 +1,106 @@
...
+				<small>
+					{updateNotification}
+					<br />
+				</small>
+			}
+		/>
+	);
+}
+
+function formatCurrentVersion(uiConfig: IPartialUiConfig): string {
```
FredrikOseberg, line 79: I usually place these in helper files inside of the component directory. It's not a must, but I feel good knowing that a component file just keeps the component and then the directory contains related functionality: 

```
-- ApiDetails
      -- ApiDetails.tsx
      -- helpers.ts
```
   

I'm open to listen to suggestions, but I feel it's sometimes easier to understand a file if I remove the clutter from the component and I find that if it's imported it's easier to locate rather than having them at the bottom of the file.

src/component/menu/ApiDetails/ApiDetails.tsx
```diff
@@ -0,0 +1,106 @@
+import useUiConfig from '../../../hooks/api/getters/useUiConfig/useUiConfig';
+import { ReactElement } from 'react';
+import ConditionallyRender from '../../common/ConditionallyRender';
+import { IVersionInfo } from '../../../interfaces/uiConfig';
+
```
FredrikOseberg, line 5: If this component is only used in the Footer, I think we can actually move it to the Footer directory and have it live under there.

FredrikOseberg (changes requested):
Looks clean! Just a few things I'd like to align on before we merge.

src/component/menu/ApiDetails/ApiDetails.tsx
olav: Good point. I do like the small helper components, but either way works for me. :)

src/component/common/ConditionallyRender/ConditionallyRender.tsx
```diff
@@ -1,5 +1,5 @@
 interface IConditionallyRenderProps {
-    condition: boolean;
+    condition: boolean | string | undefined;
```
ivarconr, line 2: I find it a bit confusing that the condition can be `boolean | string | undefined;`

src/component/menu/Footer/ApiDetails.tsx
```diff
@@ -0,0 +1,51 @@
...
+	const currentVersion = formatCurrentVersion(props.uiConfig);
+	const environment = props.uiConfig.environment;
+	const updateNotification = formatUpdateNotification(props.uiConfig);
+	
+	return (
+		<section title="API details">
+			<h4>
+				{currentVersion}{' '}
+				<ConditionallyRender
+					condition={environment}
```
ivarconr, line 29: you could automatically convert this to a boolean with the `!!`. I guess we need to decide on which pattern we prefer @FredrikOseberg 

```suggestion
					condition={!!environment}
```

FredrikOseberg: I usually prefer `Boolean(environment)` just because it's more explicit.

src/component/common/ConditionallyRender/ConditionallyRender.tsx
olav: Agreed, that is better. Fixed!

FredrikOseberg: Yeah, I think we could just have this as boolean and do the conversion. Makes it easier to reason about.

src/component/menu/Footer/ApiDetails.tsx
olav: True! I like `Boolean(foo)`. Matches `String(foo)`, `Number(foo)` etc.

src/component/menu/Footer/ApiDetails.helpers.tsx
```diff
@@ -0,0 +1,30 @@
+import { IPartialUiConfig } from './ApiDetails';
+
```
FredrikOseberg, line 2: This looks good. Just a sidenote, I usually group the child components in folders containing files pertaining to that component, and keep the helpers name lowercase (we use uppercase to distinguish component files): 

```
-- Footer
   -- Footer.tsx
   -- ApiDetails
       -- ApiDetails.tsx
       -- apidetails.helpers.ts
```

olav: Ah, that makes sense. Should we use `helpers.ts` or `api-details.helpers.ts` in this case?

FredrikOseberg: Previously I've used helpers.ts, but it can be a bit tedious if we have a lot of helper.ts files (though I've always been able to locate it by searching for the component name). I don't have strong feelings in this regard, so feel free to do what feels best to you.

FredrikOseberg: I kinda like apidetails.helpers.ts

FredrikOseberg (approved):
LGTM

PR #658, opened 2022-02-02
fix: remove projectId from create feature form

ykhedher:


ykhedher:
[bug: we should not send projectId in body when creating a feature toggle](https://trello.com/c/mn9Uhu9i/832-bug-we-should-not-send-projectid-in-body-when-creating-a-feature-toggle)

FredrikOseberg (approved):
LGTM

src/component/feature/CreateFeature/hooks/useFeatureForm.ts
```diff
@@ -42,7 +42,6 @@ const useFeatureForm = (
         return {
             type: type,
             name: name,
-            projectId: project,
```
FredrikOseberg, line 45: What do we actually use project for now?

FredrikOseberg (changes requested):
On closer inspection we need to validate that we are using the projectId from the dropdown to send the request to the correct URL.

src/component/feature/CreateFeature/hooks/useFeatureForm.ts
ivarconr: The form will probably still need to capture the projectId in order to call the correct API endpoint to create the feature toggle (projectId is part of the URL). We should just no send is as part of the json body anymore.

ykhedher: We just removed the projectId from the body payload but we still keep it in the URL.

FredrikOseberg (approved):
LGTM

PR #663, opened 2022-02-03
refactor: port FeatureToggleList to TS/SWR

olav:


src/page/archive/index.js
```diff
@@ -1,11 +1,6 @@
...
-import Archive from '../../component/archive/archive-list-container';
-import PropTypes from 'prop-types';
+import { ArchiveListContainer } from '../../component/archive/ArchiveListContainer';
 
-const render = ({ match: { params }, history }) => <Archive name={params.name} history={history} />;
-render.propTypes = {
-    match: PropTypes.object,
-    history: PropTypes.object,
-};
+const render = () => <ArchiveListContainer />;
```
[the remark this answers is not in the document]
olav: @sonatype-lift ignore

src/page/features/index.js
```diff
@@ -1,11 +1,6 @@
...
-import FeatureListContainer from '../../component/feature/FeatureToggleList';
-import PropTypes from 'prop-types';
+import { FeatureToggleListContainer } from '../../component/feature/FeatureToggleList';
 
-const render = ({ history }) => <FeatureListContainer history={history} />;
-
-render.propTypes = {
-    history: PropTypes.object.isRequired,
-};
+const render = () => <FeatureToggleListContainer />;
```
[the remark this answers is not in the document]
olav: @sonatype-lift ignore

src/hooks/useFeaturesFilter.ts
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+};
+
+const filterFeatureByRegExp = (
+    feature: IFeatureToggle,
+    filter: IFeaturesFilter,
+    regExp: RegExp
+): boolean => {
+    if (regExp.test(feature.name) || regExp.test(feature.description)) {
```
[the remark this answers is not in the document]
olav: @sonatype-lift ignore

src/hooks/api/getters/useFeatures/useFeatures.ts
```diff
@@ -1,40 +1,39 @@
...
-
-    const FEATURES_CACHE_KEY = 'api/admin/features/';
-
-    const { data, error } = useSWR(FEATURES_CACHE_KEY, fetcher, {
-        ...options,
-    });
+const PATH = formatApiPath('api/admin/features');
 
-    const [loading, setLoading] = useState(!error && !data);
+export interface UseFeaturesOutput {
```
FredrikOseberg, line 9:
```suggestion
export interface IUseFeaturesOutput {
```

We should standardize the naming of hooks return values. I think Output works well, but let's make sure we keep it consistent when we update other hooks.

src/hooks/api/getters/useFeatures/useFeatures.ts
```diff
@@ -1,40 +1,39 @@
...
+    const { data, error } = useSWR<{ features: IFeatureToggle[] }>(
+        PATH,
+        fetchFeatures,
+        options
+    );
 
-    useEffect(() => {
-        setLoading(!error && !data);
-    }, [data, error]);
+    const refetchFeatures = useCallback(() => {
```
FredrikOseberg, line 23: There's an interesting conversation to be had here!

Is this actually faster than not using useCallback? I'm asking because I don't actually know. I read an article on this topic a year back that made me think about these kinds of optimizations. I think it was [this one](https://kentcdodds.com/blog/usememo-and-usecallback)

Now, personally I'm really into performance optimizations. I find that it is especially important with the amount of JS we are putting in the browser these days, but using these tools do add complexity so I think we should think about when and where to use them. React is so fast these days, and the v8 engine caches functions where inputs do not change regardless of what we do in react. 

If it makes sense, then we should of course add it. But we should also keep this in mind: 
(1) Not doing stuff is faster than doing stuff 
(2) You don't have a performance problem until you have a performance problem

src/page/archive/index.jsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
```
FredrikOseberg, line 1: I think we can remove these pages and use the container directly in routes.js

src/page/features/index.jsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
```
FredrikOseberg, line 1: This one too.

FredrikOseberg (approved):
Looks really clean! Just some minor comments from me.

src/hooks/api/getters/useFeatures/useFeatures.ts
olav: Yeah, I was wondering about that name. Will add the `I` for now, at least.

src/hooks/api/getters/useFeatures/useFeatures.ts
olav: Good point! I didn't really consider the local performance aspect. My only goal here was for the hook to return a stable reference to `refetchFeatures`, so that it plays well with things like `React.memo`. Always nice when hooks return stable stuff, easier for consumers.

FredrikOseberg: Ahh, that makes sense.

src/component/feature/FeatureToggleList/FeatureToggleList.jsx
```diff
@@ -23,43 +23,31 @@ import { useStyles } from './styles';
 import ListPlaceholder from '../../common/ListPlaceholder/ListPlaceholder';
 import { getCreateTogglePath } from '../../../utils/route-path-helpers';
 import { NAVIGATE_TO_CREATE_FEATURE } from '../../../testIds';
+import { resolveFilteredProjectId } from "../../../hooks/useFeaturesFilter";
```
FredrikOseberg, line 26: Hm. Why is prettier not taking effect here?

src/component/feature/FeatureToggleList/index.tsx
```diff
@@ -0,0 +1,24 @@
+import { useFeatures } from '../../../hooks/api/getters/useFeatures/useFeatures';
+import useUiConfig from '../../../hooks/api/getters/useUiConfig/useUiConfig';
+import { useFeaturesFilter } from '../../../hooks/useFeaturesFilter';
+import FeatureToggleList from './FeatureToggleList';
+import { useFeaturesSort } from '../../../hooks/useFeaturesSort';
+
+export const FeatureToggleListContainer = () => {
```
FredrikOseberg, line 7: Prefer to have the file named the same as the component instead of index.tsx, it makes it easier when searching for that component in the IDE.

src/hooks/api/getters/useFeatures/useFeatures.ts
```diff
@@ -1,40 +1,39 @@
...
-        error,
-        loading,
+        loading: !error && !data,
         refetchFeatures,
+        error,
     };
 };
 
-export default useFeatures;
+function fetchFeatures() {
```
FredrikOseberg, line 35: Could be arrow for consistency :)

But I like this pattern, we should move the fetcher and path outside of the hook in every instance where it's not dependent on input parameters.

src/hooks/api/getters/useFeaturesArchive/useFeaturesArchive.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    return {
+        archivedFeatures: data?.features || [],
+        refetchArchived,
+        loading: !error && !data,
+        error,
+    };
+};
+
+function fetchFeatures() {
```
FredrikOseberg, line 37: This one could too.

FredrikOseberg (approved):
Looks good! Only minor suggestions.

src/component/feature/FeatureToggleList/FeatureToggleList.jsx
olav: Probably my IDE using the wrong style for auto-imports. Will fix. :+1:

PR #666, opened 2022-02-03
refactor: detach ApplicationList from global settings 

olav:
Simplifies the upcoming global settings refactor.

### DIFF
--- a/src/component/application/application-list-component.js
+++ b/src/component/application/application-list-component.js
@@ -25,8 +25,7 @@ const Empty = () => (
     </React.Fragment>
 );
 
-const ClientStrategies = props => {
-    const { fetchAll, applications } = props
+const ClientStrategies = ({ fetchAll, applications }) => {
     const [filter, setFilter] = useState('');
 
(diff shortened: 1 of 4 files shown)


src/component/application/application-list-component.js
```diff
@@ -25,49 +25,46 @@ const Empty = () => (
...
 );
 
-class ClientStrategies extends Component {
-    static propTypes = {
-        applications: PropTypes.array,
-        fetchAll: PropTypes.func.isRequired,
-        settings: PropTypes.object.isRequired,
-        updateSetting: PropTypes.func.isRequired,
-    };
+const ClientStrategies = props => {
```
FredrikOseberg, line 28: Just a small note. We've kept a pattern of destructuring inline: 
```suggestion
const ClientStrategies = ({ fetchAll, applications }) => {
```

Would be good to keep the consistency, and it also makes it easy to see component dependencies at a glance.

FredrikOseberg (approved):
LGTM. Just a small note.

PR #668, opened 2022-02-03
Refactor/strategies

ykhedher:


ykhedher:
[Refactor redux - Strategies](https://trello.com/c/qtmO8LpO/823-refactor-redux-strategies)

src/component/strategies/ShowStrategy.tsx
```diff
@@ -0,0 +1,85 @@
+import {
```
FredrikOseberg, line 1: We need to land a naming convention on these views. Right now we have too many different ones. I would like us to either use StrategyView or something similar for now.

src/component/strategies/StrategiesList/StrategiesList.tsx
```diff
@@ -99,16 +97,70 @@ const StrategiesList = ({
         </Link>
     );
 
+    const onReactivateStrategy = strategy => {
+        setDialogueMetaData({
+            show: true,
+            title: 'Really reactivate strategy?',
+            onConfirm: () => {
```
FredrikOseberg, line 104: Remember this needs to await.

FredrikOseberg (approved):
LGTM. Some TS to clean up later, but the important part is done!

PR #669, opened 2022-02-03
refactor: port EventHistory to TS/SWR

olav:


src/page/history/toggle.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import HistoryListToggle from '../../component/history/FeatureEventHistory';
+import { FeatureEventHistory } from '../../component/history/FeatureEventHistory/FeatureEventHistory';
 
 const render = ({ match: { params } }) => (
-    <HistoryListToggle toggleName={params.toggleName} />
+    <FeatureEventHistory toggleName={params.toggleName} />
```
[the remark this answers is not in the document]
olav: @sonatype-lift ignore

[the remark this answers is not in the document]
olav: @sonatype-lift ignore

src/hooks/api/getters/useEvents/useEvents.ts
```diff
@@ -0,0 +1,52 @@
+import useSWR, { mutate, SWRConfiguration } from 'swr';
+import { useCallback } from 'react';
+import { formatApiPath } from '../../../../utils/format-path';
+import handleErrorResponses from '../httpErrorResponseHandler';
+import { IEvent } from '../../../../interfaces/event';
+
+const PATH = formatApiPath('api/admin/events');
+
+export interface UseEventsOutput {
```
FredrikOseberg, line 9:
```suggestion
export interface IUseEventsOutput {
```

src/hooks/api/getters/useEvents/useEvents.ts
```diff
@@ -0,0 +1,52 @@
...
+    events: IEvent[];
+    refetchEvents: () => void;
+    loading: boolean;
+    error?: Error;
+}
+
+export const useEvents = (
```
FredrikOseberg, line 16: I'm wondering if this could be easier represented as two separate hooks: 

```
useEvents
useFeatureEvents
```

I think what you have done here is cool, but I had to spend some time going through what was happening. It would be easier to reason about if they were separate hooks, and perhaps also easier to understand what was happening in the context they were being used. What do you think?

src/hooks/api/getters/useEvents/useEvents.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    return {
+        events: data?.events || [],
+        loading: !error && !data,
+        refetchEvents,
+        error,
+    };
+};
+
+function fetchEvents(featureName?: string) {
```
FredrikOseberg, line 38: We've tried to use arrow functions as much as possible for new code, just to keep consistency and eventually reach a state where things are similar enough that it reduces the cognitive load.

FredrikOseberg (changes requested):
Nice work! Just want to align some minor things before we merge.

src/hooks/api/getters/useEvents/useEvents.ts
olav: Yeah, that's much better! Will fix.

FredrikOseberg (approved):
Nice!

PR #674, opened 2022-02-04
refactor: port ReportCard to TS/SWR

olav:


src/interfaces/project.ts
```diff
@@ -24,7 +24,7 @@ export interface IProjectHealthReport extends IProject {
     staleCount: number;
     potentiallyStaleCount: number;
     activeCount: number;
-    updatedAt: Date;
+    updatedAt: string;
```
ykhedher, line 27: is it really a type of string ?

ykhedher (approved):
Looks good for me.

src/interfaces/project.ts
olav: Yep! This type describes what we get from the backend.

FredrikOseberg (approved):
LGTM

PR #675, opened 2022-02-07
refactor: port auth admin to TS/SWR

olav:


src/component/admin/auth/AutoCreateForm/AutoCreateForm.tsx
```diff
@@ -1,102 +1,120 @@
...
         autoCreate: boolean;
         defaultRootRole?: string;
         emailDomains?: string;
-
     };
     setValue: (name: string, value: string | boolean) => void;
 }
 
-function AutoCreateForm({ data = { enabled: false, autoCreate: false }, setValue }: Props) {
+function AutoCreateForm({
```
FredrikOseberg, line 23: Small detail, but perhaps we could change this to an arrow function and explicit export?

FredrikOseberg (approved):
Looks good! I have some small comments and suggestions. Some of which are just personal preference. 

As a sidenote, I think we should consider covering this with an e2e test. But first we need to setup an enterprise test server.

PR #679, opened 2022-02-08
refactor: port global settings to TS/hooks

olav:


src/component/admin/users/UsersList/UsersList.tsx
```diff
@@ -104,7 +103,7 @@ const UsersList = () => {
                     user={user}
                     openPwDialog={openPwDialog}
                     openDelDialog={openDelDialog}
-                    location={location}
+                    location={locationSettings}
```
FredrikOseberg, line 106: I think we should try to keep the props consistent aswell through components. Tracking the data flow gets trickier if the name of the property changes through the component tree.

src/component/history/EventLog/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useLocationSettings } from "../../../hooks/useLocationSettings";
+
+interface IEventLogContainerProps {
+    title: string;
+    history: unknown[];
+    displayInline?: boolean;
+}
+
+const EventLogContainer = (props: IEventLogContainerProps) => {
+    const [locationSettings] = useLocationSettings();
```
FredrikOseberg, line 12: Is there a significant benefit here of using an array instead of an object? I know we can't keep it 100% consistent across all hooks, but if we can land a pattern for our own hooks then it will help reducing the cognitive load (types will also help). I think almost all of our hooks currently return objects.

FredrikOseberg (approved):
Another sound PR. Just minor suggestions that could align some more consistency.

src/component/history/EventLog/index.tsx
olav: Good point. I was mimicking `useState`, but the object-form would enforce consistent names. Will fix.

FredrikOseberg: I don't blame you! I think it's sensible in a lot of situations, but I've found that the array form sometimes can be a bit limiting especially when returning more than two values - or when you just need the second item in the array, resulting in:
```
[, setSomething] = useSomething();
``` 

With setState it's not that much of an issue since you'll always need the piece of state and setter (otherwise you wouldn't need state)

PR #680, opened 2022-02-08
refactor: port login auth to TS/SWR

olav:


src/component/App.tsx
```diff
@@ -1,58 +1,32 @@
...
-    // because we need the userId when the component load.
-    const { splash, user: userFromUseUser, authDetails } = useUser();
 
-    const [showSplash, setShowSplash] = useState(false);
-    const [showLoader, setShowLoader] = useState(false);
-    useEffect(() => {
-        fetchUiBootstrap();
-        /* eslint-disable-next-line */
-    }, [user.authDetails?.type]);
+export const App = ({ fetchUiBootstrap }: IAppProps) => {
```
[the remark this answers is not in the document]
olav: Nice catch. We can take a look at `fetchUIBootstrap` once this PR is in.

src/component/tag-types/form-tag-type-component.js
```diff
@@ -4,15 +4,12 @@ import classnames from 'classnames';
 
 import { FormButtons } from '../common';
 import PageContent from '../common/PageContent/PageContent';
-import { Typography, TextField } from '@material-ui/core';
+import { TextField, Typography } from '@material-ui/core';
 
 import styles from './TagType.module.scss';
 import commonStyles from '../common/common.module.scss';
```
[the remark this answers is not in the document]
olav: Yeah, I've deleted these in an upcoming PR: https://github.com/Unleash/unleash-frontend/pull/687

src/component/App.tsx
```diff
@@ -1,58 +1,32 @@
...
+    const { auth, refetchAuth } = useAuth();
```
FredrikOseberg, line 22: I'm not sure how we should name this. This name suggests that it contains authentication details, but it's actually encompassing different data at different points in time: 

(1) If the user is not authenticated, the call will return authDetails which is a set of data used to render the different login options in the login page
(2) If the user is authentication, then the API endpoint returns user _metadata_ and user _authorization_ in the form of permission. 

I think this might be a flaw in our API design, but perhaps we could alleviate this a bit by creating abstractions on top of the base useSWR hook. 

I'm thinking something along the lines of: 

```
useUserEndpoint -> Base hook
useAuthentication -> Use the useUserEndpoint and expose authDetails
useUser -> Use the useUserEndpoint and expose user details
useAuthorization -> Use the useUserEndpoint and expose user permissions
```

Since all hooks will use the same SWR cache there should be no degradation of performance doing it this way, and it might be easier to understand in different contexts of the code.  Thoughts?

src/component/providers/AccessProvider/AccessProvider.tsx
```diff
@@ -1,88 +1,110 @@
...
-        let result = false;
-
-        return () => {
-            if (called) return result;
-            const permissions = store.getState().user.get('permissions') || [];
-            result = permissions.some(
-                (p: IPermission) => p.permission === ADMIN
+const AccessProvider = (props: IAccessProviderProps): ReactElement => {
+    const { auth } = useAuth();
+    const permissions = props.permissions ?? auth?.permissions;
```
FredrikOseberg, line 14: Is there any benefit of receiving this from props rather than useAuth? I imagine the source would be the same for both of these.

src/component/providers/AccessProvider/AccessProvider.tsx
```diff
@@ -1,88 +1,110 @@
...
+    const hasAccess = useCallback(
+        (permission: string, project?: string, environment?: string) => {
+            return checkPermissions(
+                permissions,
+                permission,
+                project,
+                environment
             );
+        },
+        [permissions]
```
FredrikOseberg, line 29: Given recent discussions, perhaps we should change this to JSON.stringify(permissions) or use deep comparison hook? As far as I know useCallback also does a shallow comparison.

src/component/user/DemoAuth/DemoAuth.jsx
```diff
@@ -1,25 +1,24 @@
...
+    const { emailAuth } = useAuthApi();
 
     const handleSubmit = evt => {
         evt.preventDefault();
-        const user = { email };
-        const path = evt.target.action;
 
-        demoLogin(path, user).then(() => {
-            refetch();
+        emailAuth(authDetails.path, email).then(() => {
```
FredrikOseberg, line 20: I think we should try to land a pattern here too. Most other parts use try / catch.

src/hooks/api/actions/useAuthApi/useAuthApi.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+    return {
+        passwordAuth,
+        emailAuth,
+        errors,
+        loading,
+    };
+};
+
+const ensureRelativePath = (path: string): string => {
```
FredrikOseberg, line 49: Cool, if we move this into the createRequest method this will apply to all hooks on the API side at least.

FredrikOseberg (changes requested):
Great start. Some things I'd like to align on, and still need to perform some testing on this.

src/component/providers/AccessProvider/AccessProvider.tsx
olav: Yeah, the prop is only used for testing. Mocking `useAuth` would probably be better, but I think that should be its own PR.

FredrikOseberg: Makes sense! Maybe add a comment so we remember?

src/component/providers/AccessProvider/AccessProvider.tsx
olav: The `permissions` list from `useAuth` should be a stable reference now (through `useMemo`), but a good idea to double check. :eyes:

src/component/App.tsx
olav: Good idea! Pushed a change which splits the hooks. 

Still not sure about the naming, but I like that they share a prefix since they fetch from the same endpoint.

src/hooks/api/actions/useAuthApi/useAuthApi.tsx
olav: Agreed, maybe even into `formatApiPath` and `formatAssetPath`. Should probably be a separate PR.

cypress/integration/auth/auth.spec.js
```diff
@@ -0,0 +1,160 @@
...
+// for a todo list app that are designed to demonstrate
+// the power of writing tests in Cypress.
+//
+// To learn more about how Cypress works and
+// what makes it such an awesome testing tool,
+// please read our getting started guide:
+// https://on.cypress.io/introduction-to-cypress
+
+const username = 'test@test.com';
+const password = 'qY70$NDcJNXA';
```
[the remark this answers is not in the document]
FredrikOseberg: ignore

cypress/integration/feature-toggle/auth.spec.js
```diff
@@ -0,0 +1,60 @@
...
+// This spec file contains a variety of sample tests
+// for a todo list app that are designed to demonstrate
+// the power of writing tests in Cypress.
+//
+// To learn more about how Cypress works and
+// what makes it such an awesome testing tool,
+// please read our getting started guide:
+// https://on.cypress.io/introduction-to-cypress
+
+const username = 'test@test.com';
```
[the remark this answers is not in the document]
FredrikOseberg: ignore

cypress/integration/feature-toggle/auth.spec.js
```diff
@@ -0,0 +1,60 @@
...
+const password = 'qY70$NDcJNXA';
```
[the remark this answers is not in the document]
FredrikOseberg: @sonatype-lift ignore

PR #682, opened 2022-02-09
refactor: remove unused feedback state

olav:


src/component/common/util.js
```diff
@@ -114,32 +113,3 @@ export const modalStyles = {
...
-            feedback => feedback.feedbackId === 'pnps'
-        );
-
-        if (!feedback) return false;
-
-        if (feedback.neverShow) {
-            return false;
-        }
-
-        if (feedback.given) {
```
FredrikOseberg, line 135: I think I have forgotten to add this check to the Feedback component itself when refactoring. The intention is that if you have given feedback we should start showing the component again after six months. I think it's current state is to return null if given is true. Could we add that to this PR?

FredrikOseberg (approved):
LGTM. The feedback component was originally intended to be a multipurpose component, but I realize that it's going to take some greater refactoring for it to be that at this point, and it's just used for pnps at this point - so let's keep it like this and refactor later if we need it for anything else.

PR #684, opened 2022-02-09
refactor: port MainLayout to TS/SWR 

olav:


src/component/layout/MainLayout/MainLayout.tsx
```diff
@@ -21,24 +20,30 @@ const useStyles = makeStyles(theme => ({
         height: '100%',
         padding: '3.25rem 0',
         position: 'relative',
+        // @ts-expect-error
```
FredrikOseberg, line 23: Here's a subtle trick. If you change the import from `import { makeStyles } from '@material-ui/styles';` to: `import { makeStyles } from '@material-ui/core/styles';`

It will associate the correct types and you won't need the @ts-expect-error

olav: TIL!

PR #691, opened 2022-02-09
poc: use unleash

FredrikOseberg:
https://www.loom.com/share/70da1de4a8da4098a93d9ce77f855c11

First attempt at integrating unleash in non-invasive way for OSS.


FredrikOseberg:
> I guess the thought here is to be bring this in so users can test their feature toggles in the GUI?

That's another use case I hadn't even thought about. But it will in fact also enable that, yeah!

PR #699, opened 2022-02-11
fix: add label for the impression data

ykhedher:


src/component/feature/FeatureForm/FeatureForm.tsx
```diff
@@ -157,12 +160,20 @@ const FeatureForm: React.FC<IFeatureToggleForm> = ({
...
+                            control={
+                                <Switch
+                                    name="impressionData"
+                                    onChange={() =>
+                                        setImpressionData(!impressionData)
+                                    }
+                                    checked={impressionData}
+                                />
+                            }
+                            label="Impression Data"
```
thomasheartman, line 175: Nice! 🥳

thomasheartman (changes requested):
Nice one! 

Just a little thought. Now we have two labels that say "Impression Data" (sure, the top one is technically not a `label`, but it looks the same):

image

Do we want that? It feels a bit confusing.

How about changing the `Typography` element on lines 141-148 to a `FormControlLabel` and provide it a reference to the Impression data switch? That'll both cover the label duplication _and_ it feels more semantically correct to me (though admittedly, I may be off on this one) 💁🏼

ykhedher:
[Impression data input is unlabeled](https://trello.com/c/Sd8Ubdfj/861-impression-data-input-is-unlabeled)

thomasheartman (approved):
Yeah, that's much better 😄 🙌🏼  Good work!

image

PR #704, opened 2022-02-11
refactor: use copy library and add explicit export

ykhedher:


ykhedher:
[Port UserToken to use 'copy-to-clipboard' library](https://trello.com/c/5AFm9t4v/869-port-usertoken-to-use-copy-to-clipboard-library)

FredrikOseberg (approved):
LGTM.

PR #708, opened 2022-02-14
fix: remove disabled check

FredrikOseberg:


olav (approved):

FredrikOseberg:
[bug: cannot disable default environment.](https://trello.com/c/pf18g76m/876-bug-cannot-disable-default-environment)

PR #709, opened 2022-02-14
refactor: fix misc test feedback

olav:


src/utils/truncate-string.test.ts
```diff
@@ -0,0 +1,9 @@
+import { truncateString } from './truncate-string';
```
FredrikOseberg, line 1: Nice util. We have a utility component called StringTruncator that you can give a max width and it will wrap the text with a tooltip that you can hover to view the full name. It's not perfect but I think it would be nice if we could display that information too.

FredrikOseberg (changes requested):
Could we look into using or modifying the StringTruncator to be used in these cases too? For resources where the name is important (such as environments and project ids), we should be able to display that information by hovering.

src/component/addons/AddonList/ConfiguredAddons/ConfiguredAddons.tsx
```diff
@@ -127,7 +137,7 @@ export const ConfiguredAddons = ({ getAddonIcon }: IConfigureAddonsProps) => {
                         setShowDelete(true);
                     }}
                 >
-                    <Delete />
+                    <Delete titleAccess="Remove Addon" />
```
FredrikOseberg, line 140: Nice.

src/component/feature/FeatureView/FeatureOverview/StaleDialog/StaleDialog.tsx
```diff
@@ -30,12 +33,26 @@ const StaleDialog = ({ open, setOpen, stale }: IStaleDialogProps) => {
 
     const toggleActionText = stale ? 'active' : 'stale';
 
-    const onSubmit = async e => {
-        e.stopPropagation();
+    const onSubmit = async (event: React.SyntheticEvent) => {
+        event.stopPropagation();
         const patch = [{ op: 'replace', path: '/stale', value: !stale }];
         await patchFeatureToggle(projectId, featureId, patch);
```
FredrikOseberg, line 39: Hmm, this wasn't done in this PR. But it seems to be an oversight that this is not wrapped in a try catch?

src/component/strategies/StrategyForm/StrategyForm.tsx
```diff
@@ -78,34 +79,29 @@ export const StrategyForm = ({ editMode, strategy }: IStrategyFormProps) => {
...
                 setToastData({
                     type: 'success',
                     title: 'Success',
                     text: 'Successfully updated strategy',
                 });
                 refetchStrategies();
-            } catch (e: any) {
-                setToastApiError(e.toString());
+            } catch (error: unknown) {
+                setToastApiError(formatUnknownError(error));
```
FredrikOseberg, line 91: Nice.

src/component/tags/TagTypeForm/useTagTypeForm.ts
```diff
@@ -39,14 +38,12 @@ const useTagTypeForm = (initialTagName = '', initialTagDesc = '') => {
...
-                setErrors(prev => ({
-                    ...prev,
-                    name: NAME_EXISTS_ERROR,
-                }));
-                return false;
-            }
+        } catch (err: unknown) {
+            setErrors(prev => ({
+                ...prev,
+                name: formatUnknownError(err)
```
FredrikOseberg, line 44: Let's make a note of updating this elsewhere in the application aswell.

FredrikOseberg (approved):
Nice. This is almost good to go, let's just check that the patch call is properly error handled before we merge.

PR #730, opened 2022-02-21
feat: add redirect params to /login

ykhedher:


src/component/common/ProtectedRoute/ProtectedRoute.jsx
```diff
@@ -1,17 +1,21 @@
...
+    const history = useHistory();
+    const loginLink =
+        pathname.length > 1 ? `/login?redirect=${pathname}` : '/login';
     return (
         <Route
             {...rest}
             render={props => {
                 if (unauthorized) {
-                    return <Redirect to={'/login'} />;
+                    history.push(loginLink);
```
olav, line 18: Could keep the Redirect component here, so that we return `null` and `replace` the current path:

```
return <Redirect to={loginLink} />;
```

src/component/user/Login/Login.tsx
```diff
@@ -7,17 +7,19 @@ import { DEMO_TYPE } from '../../../constants/authTypes';
...
     const { authDetails } = useAuthDetails();
     const { user } = useAuthUser();
     const query = useQueryParams();
     const resetPassword = query.get('reset') === 'true';
+    const redirect = query.get('redirect') || '/features';
+    const history = useHistory();
 
     if (user) {
-        return <Redirect to="/features" />;
+        history.replace(redirect);
```
olav, line 22: Same as above, should keep `return` and `Redirect`.

src/component/common/ProtectedRoute/ProtectedRoute.jsx
```diff
@@ -1,17 +1,21 @@
...
 const ProtectedRoute = ({
     component: Component,
     unauthorized,
     renderProps = {},
     ...rest
 }) => {
+    const { pathname } = useLocation();
+    const history = useHistory();
+    const loginLink =
+        pathname.length > 1 ? `/login?redirect=${pathname}` : '/login';
```
olav, line 12: Could we always add the param? Should test this both with and without a baseUriPath.

src/component/user/DemoAuth/DemoAuth.jsx
```diff
@@ -9,7 +9,7 @@ import { useAuthApi } from '../../../hooks/api/actions/useAuthApi/useAuthApi';
 import { useAuthUser } from '../../../hooks/api/getters/useAuth/useAuthUser';
 import useToast from '../../../hooks/useToast';
 
-const DemoAuth = ({ authDetails }) => {
+const DemoAuth = ({ authDetails, redirect }) => {
```
olav, line 12: Add to propTypes.

olav (changes requested):

src/component/user/HostedAuth/HostedAuth.jsx
```diff
@@ -18,7 +18,7 @@ import {
     LOGIN_PASSWORD_ID,
 } from '../../../testIds';
 
-const HostedAuth = ({ authDetails }) => {
+const HostedAuth = ({ authDetails, redirect }) => {
```
olav, line 21: Should add this to propTypes as well. Same for the other Auth components.

src/component/user/Login/Login.tsx
```diff
@@ -7,17 +7,18 @@ import { DEMO_TYPE } from '../../../constants/authTypes';
...
 const Login = () => {
     const styles = useStyles();
     const { authDetails } = useAuthDetails();
     const { user } = useAuthUser();
     const query = useQueryParams();
     const resetPassword = query.get('reset') === 'true';
+    const redirect = query.get('redirect') || '/';
 
     if (user) {
-        return <Redirect to="/features" />;
```
olav, line 20: I suppose we should still redirect to the features page on login. Does that happen somewhere else?

src/component/common/ProtectedRoute/ProtectedRoute.jsx
olav: Should we remove the length condition, and always add the query param? Seems simpler.

olav (changes requested):

src/component/user/Login/Login.tsx
ykhedher: Yes. so if there's no redirect we will get the `/` and this is matched in Apps.ts L:85
```
<ProtectedRoute
    exact
    path="/"
    unauthorized={isUnauthorized()}
    component={Redirect}
    renderProps={{ to: '/features' }}
/>
```

olav (approved):

src/component/user/Login/Login.tsx
```diff
@@ -15,9 +15,10 @@ const Login = () => {
     const { user } = useAuthUser();
     const query = useQueryParams();
     const resetPassword = query.get('reset') === 'true';
```
FredrikOseberg, line 17: Where does this get set, and will it overwrite or add to the query params? Worst case we will redirect to "/", but would be nice to check what happens.

FredrikOseberg (approved):
Think this looks good. Since we are using internal react-router methods the basename will handle the prefix. Can we test this somehow?

src/component/user/Authentication/Authentication.tsx
```diff
@@ -15,7 +15,10 @@ import ConditionallyRender from '../../common/ConditionallyRender';
 import { Alert } from '@material-ui/lab';
 import { useAuthDetails } from '../../../hooks/api/getters/useAuth/useAuthDetails';
 
-const Authentication = () => {
+interface IAuthenticationProps {
+    redirect: string;
+}
+const Authentication = ({ redirect }: IAuthenticationProps) => {
```
ivarconr, line 21: We need to validate the redirect to avoid the unvalidated redirect vulnerability.

https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html

ivarconr (changes requested):
We need to validate the redirect to avoid the unvalidated redirect vulnerability.

https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html

src/component/user/Authentication/Authentication.tsx
ykhedher: @ivarconr I think that we are good as we are redirected to 404 page if we put a link not existing. Even if we try to redirect to external websites it won't work.
Also I did test it with subpath: https://www.loom.com/share/262a82afd4ad4e5b9581331bb15286f8 and it is working ✅

ivarconr: perfect, thanks!

ivarconr (approved):

PR #737, opened 2022-02-22
refactor: make text in ui consistent

ykhedher:


ykhedher:
[Make text in the UI consistent: 'new x', 'create x', 'save' etc ...](https://trello.com/c/Z8M4Bxjl/890-make-text-in-the-ui-consistent-new-x-create-x-save-etc)

thomasheartman (changes requested):
Nice work! 🙌🏼  I've got a suggestion, though. Can we abstract this into reusable components? E.g. have
- a `ResourceCreationButton` that accepts the `resourceType` (and other necessary props)? 
- a `SaveChangesButton` that accepts `permission`
- and any other changes that are in the same vein?

That way, it would be much easier to keep in sync in the future 🤓 

What do you think?

ykhedher:
> Nice work! 🙌🏼 I've got a suggestion, though. Can we abstract this into reusable components? E.g. have
> 
>     * a `ResourceCreationButton` that accepts the `resourceType` (and other necessary props)?
> 
>     * a `SaveChangesButton` that accepts `permission`
> 
>     * and any other changes that are in the same vein?
> 
> 
> That way, it would be much easier to keep in sync in the future 🤓
> 
> What do you think?

@thomasheartman  Makes sense, I removed all the tooltips in this PR for the buttons, but for the `ResourceCreationButton` and `SaveChangesButton` I will create it in a separate PR.

thomasheartman (approved):
@ykhedher Aight, that's fine by me 🙆🏼  In that case I'll go ahead and approve this one 😄

PR #738, opened 2022-02-23
fix: remove toast when clicking cancel

ykhedher:


ykhedher:
[UI Inconsistencies](https://trello.com/c/cvM0vsCE/898-ui-inconsistencies)

src/component/addons/AddonForm/AddonForm.jsx
```diff
@@ -149,7 +149,7 @@ export const AddonForm = ({ editMode, provider, addon, fetch }) => {
                 </a>
                 <p className={commonStyles.error}>{errors.general}</p>
             </section>
-            <form onSubmit={onSubmit}>
```
olav, line 152: Instead of removing the `onSubmit` handler, maybe adding `type="button"` to the cancel button would work.

(That should stop the cancel button from submitting the form, while still allowing for regular form submissions.)

olav (changes requested):

olav (approved):

src/component/addons/AddonForm/AddonForm.jsx
```diff
@@ -79,7 +79,7 @@ export const AddonForm = ({ editMode, provider, addon, fetch }) => {
         setErrors({ ...errors, events: undefined });
     };
 
-    const handleCancel = () => {
+    const onCancel = () => {
```
FredrikOseberg, line 82: There are some issues with using `history.goBack()`. It works fine if you always think that the user will get to this page by usual flow. But if you go to this page via a direct link, then the cancel button will suddently take you back to the page in the history where you came from. 

Try going directly here and hit cancel: https://unleash-frontend-git-fix-remove-toast-on-cancel-unleash-team.vercel.app/addons/create/slack

@olav @ykhedher WDYT? It is a low risk, but maybe it is better to use explicit paths?

FredrikOseberg (approved):
Looks good. The comment about `history.goBack` is a separate discussion.

src/component/addons/AddonForm/AddonForm.jsx
olav: Agreed, paths would be better. :+1:

PR #739, opened 2022-02-23
feat: add new feature strategy create/edit pages

olav:


src/component/feature/FeatureStrategy/FeatureStrategyEdit/FeatureStrategyEdit.tsx
```diff
@@ -0,0 +1,150 @@
...
+                formatUpdateStrategyApiCode(
+                    projectId,
+                    featureId,
+                    environmentId,
+                    strategy,
+                    unleashUrl
+                )
+            }
+        >
+            <FeatureStrategyContainer
```
FredrikOseberg, line 79: In previous form we've used the convention of ending with Form and used hooks to share functionality. Could we do the same here?

src/component/feature/FeatureStrategy/FeatureStrategyContainer/FeatureStrategyContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    />
+                )}
+                <FeatureStrategyProdGuard
+                    open={showProdGuard}
+                    onClose={() => setShowProdGuard(false)}
+                    onClick={onSubmit}
+                    loading={loading}
+                    label="Save strategy"
+                />
+                <Button
```
FredrikOseberg, line 128: Having the buttons here places them directly underneath the form instead of at the bottom like FormTemplate provider does. It would be good if we could use the same location so that if we decide to do something with the buttons (for example make them floating) we'd only have to work with the container in the FormTemplate.

FredrikOseberg (changes requested):
A few comments on the code I'd like to clear. In terms of behavior: 
- I think the Add strategy button on the accordion should remain. It makes it a lot easier to add strategies without having to open the accordion
- Constraints added should always open in edit mode
- We'll need to expose the validations from the constraints - you shouldn't be able to save strategies with invalid constraints (currently you can save strategies with constraints that have 0 values)
- I think we should add capabilities to edit multiple constraints
- Can we put groupId and stickiness in an advanced accordion? 
- Should we change the gear icon to an edit icon, and add the capability to delete a strategy directly from the accordion view?

src/component/common/ConstraintAccordion/ConstraintAccordionEdit/ConstraintAccordionEdit.tsx
```diff
@@ -183,16 +182,18 @@ export const ConstraintAccordionEdit = ({
     }, [localConstraint.operator, localConstraint.contextName, setError]);
 
     return (
-        <form onSubmit={onSubmit} className={styles.form}>
```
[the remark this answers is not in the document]
FredrikOseberg: Well. I get why it was changed, otherwise we'd get a nested form.

src/component/feature/FeatureStrategy/FeatureStrategyEdit/FeatureStrategyEdit.tsx
olav: Good point, that's a better name. I don't think there's much more to share between Create and Edit (other than what's already shared through `FeatureStrategyForm`).

src/component/common/SidebarModal/SidebarModal.tsx
```diff
@@ -0,0 +1,45 @@
...
+            open={open}
+            onClose={onClose}
+            closeAfterTransition
+            aria-label={label}
+            BackdropComponent={Backdrop}
+            BackdropProps={{ timeout: TRANSITION_DURATION }}
+        >
+            <Fade timeout={TRANSITION_DURATION} in={open}>
+                <div>
+                    <Slide
```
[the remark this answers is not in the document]
FredrikOseberg: In addition, when clicking edit strategy it fades in instead of sliding. I would prefer the behaviors to be the same.

src/component/feature/FeatureStrategy/FeatureStrategyConstraints2/FeatureStrategyConstraints2.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const stopEditingIndex = (index: number) => {
+        setEditIndexes(prev => prev.filter(editIndex => editIndex !== index));
+    };
+
+    const isEditingIndex = (index: number) => {
+        return editIndexes.includes(index);
+    };
+
+    const onAddConstraint = () => {
+        if (mode === 'create') {
```
FredrikOseberg, line 51: This is almost accurate. If you add a new strategy in edit mode, this should also start in "Editing mode".

FredrikOseberg: In addition, if you hit cancel on a new constraint (either in create or edit mode) without first saving the draft, it should remove the constraint from the list. This was the reasoning behind the metadata in the previous implementation.

src/component/feature/FeatureStrategy/FeatureStrategyConstraints2/useConstraintsValidation.ts
```diff
@@ -0,0 +1,30 @@
+import useFeatureApi from 'hooks/api/actions/useFeatureApi/useFeatureApi';
+import { useEffect, useState } from 'react';
+import { IConstraint } from 'interfaces/strategy';
+
+export const useConstraintsValidation = (
+    projectId: string,
+    featureId: string,
+    constraints?: IConstraint[]
+): boolean => {
+    const [valid, setValid] = useState(constraints?.length === 0);
```
FredrikOseberg, line 10: Are the constraints automatically valid if the list is greater than one? I guess this is for the editing mode. Would it be simpler to return an async function that accepts a list of constraints and essentially does what is done in the useEffect?

src/component/feature/FeatureStrategy/FeatureStrategyForm/FeatureStrategyForm.tsx
```diff
@@ -0,0 +1,158 @@
...
+    environmentId,
+    onSubmit,
+    loading,
+    hasAccess,
+    mode,
+}: IFeatureStrategyFormProps) => {
+    const styles = useStyles();
+    const [showProdGuard, setShowProdGuard] = useState(false);
+    const enableProdGuard = useFeatureStrategyProdGuard(feature, environmentId);
+    const StrategyIcon = getFeatureStrategyIcon(strategy.name ?? '');
```
FredrikOseberg, line 55: Hmm. I think this function is returning the wrong icons. We added a different icon for gradual rollout a while back
Skjermbilde 2022-03-08 kl  14 47 27
.

src/component/feature/FeatureStrategy/FeatureStrategyForm/FeatureStrategyForm.tsx
```diff
@@ -0,0 +1,158 @@
...
+                    }
+                />
+                <FeatureStrategyProdGuard
+                    open={showProdGuard}
+                    onClose={() => setShowProdGuard(false)}
+                    onClick={onSubmit}
+                    loading={loading}
+                    label="Save strategy"
+                />
+                <Button
```
FredrikOseberg, line 146: We might want to have this as a PermissionButton (in case someone is linking directly to this page).

FredrikOseberg (changes requested):
Some minor things to fix, but overall this looks really solid and we're close to a merge here.

src/component/feature/FeatureStrategy/FeatureStrategyConstraints2/useConstraintsValidation.ts
olav: The idea is that an empty list of constraints is valid, while a non-empty list has to be checked. Added a comment.

The useState/useEffect handles running validations whenever constraints change, storing the latest result.

src/component/common/SidebarModal/SidebarModal.tsx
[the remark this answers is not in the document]
olav: The slide-out is a bit of work since we control the modal state through routing and path parameters.

I've removed the slide-in at least, so that the animation is consistent.

src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCard/FeatureStrategyMenuCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+        strategy.name
+    );
+
+    return (
+        <Link to={createStrategyPath} className={styles.card}>
+            <div className={styles.icon}>
+                <StrategyIcon aria-hidden />
+            </div>
+            <div>
+                <div className={styles.name}>
```
FredrikOseberg, line 40: We should probably use `<StringTruncator />` here: 
Skjermbilde 2022-03-09 kl  11 36 17

src/component/feature/FeatureStrategy/FeatureStrategyForm/FeatureStrategyForm.tsx
```diff
@@ -0,0 +1,163 @@
...
+        : FeatureStrategyConstraints;
+    const disableSubmitButtonFromConstraints = uiConfig.flags.CO
+        ? !hasValidConstraints
+        : false;
+
+    return (
+        <form className={styles.form} onSubmit={onSubmitOrProdGuard}>
+            <h2 className={styles.title}>
+                <StrategyIcon className={styles.icon} aria-hidden />
+                <span className={styles.name}>{strategyName}</span>
```
FredrikOseberg, line 90: We might also want StringTruncator here.

FredrikOseberg (changes requested):
I think this is super close. Just noted a few things I found during testing, also it looks like the failing cypress test is legitimate?

PR #748, opened 2022-02-24
fix: do not display feature not found indiscriminately when an API call fails

ykhedher:


ykhedher:
[Frontend - Do not display feature not found indiscriminately when an API call fails.](https://trello.com/c/zIjNXXNV/899-frontend-do-not-display-feature-not-found-indiscriminately-when-an-api-call-fails)

src/component/feature/FeatureView/FeatureView.tsx
```diff
@@ -132,127 +132,117 @@ const FeatureView = () => {
         );
     };
 
+    if (error) {
```
FredrikOseberg, line 135: Doesn't this actually do exactly the same as conditionally render? The point is that we only want to show this if we get a 404 not found error back from the API.

FredrikOseberg (changes requested):
Might need some changes before it's good to go.

src/component/feature/FeatureView/FeatureView.tsx
```diff
@@ -133,6 +133,9 @@ export const FeatureView = () => {
         );
     };
 
+    if (error?.status === 404) {
+        return renderFeatureNotExist();
```
FredrikOseberg, line 137: We also need to check if the feature exists in archive here. Because the feature might exist and be archived.

FredrikOseberg (changes requested):

src/component/feature/FeatureView/FeatureView.tsx
ykhedher: @FredrikOseberg Done ✅

ykhedher:
#774 ✅

PR #772, opened 2022-03-08
fix: move unsaved label below refresh popup

olav:


thomasheartman (commented):
Looks good to me! Any chance it'll conflict with anything else now? Do we have a z-index overview anywhere or some other way to know what layer to assign something if anything?

olav:
@thomasheartman Good point, definitely a chance. I don't think we have named layers or anything like that, but lowering the values looks pretty safe.

FredrikOseberg (approved):
LGTM

PR #780, opened 2022-03-10
feat: add segments

ykhedher:


ykhedher:
[Segments frontend](https://trello.com/c/FG54gVeQ/926-segments-frontend)

ykhedher:
[Frontend - Create segment form](https://trello.com/c/WOcP0TIH/941-frontend-create-segment-form)

FredrikOseberg (changes requested):
I haven't had time to look through everything, but here are some superficial functionality issues I found: 
- The searchfield only allows you to select a context field _once_. What if you want to have a constraint two times with different operators?
- You can save a segment even though constraint validation isn't passing
- If you save the constraints, and go back, then forward. All constraints will open in edit mode. 
- If you hit cancel on a new constraint field, it should be removed from the list if it's a new constraint and you haven't saved it yet.

FredrikOseberg:
This text should also be reviewed: 

```
Segment makes it easy for you to define who should be exposed to your feature. The segment is often a collection of constraints and can be reused. Create a segment and use it when you configure an activation strategy on a feature toggle.
```

* A segment is _always_ a collection of constraints, not _often_.

Maybe involve @thomasheartman and Elise/Nicolae here.

FredrikOseberg:
Skjermbilde 2022-03-21 kl  16 21 36

Skjermbilde 2022-03-21 kl  16 21 45

Segments are currently listed twice in the sidebar menu (also validate that this does not render when the segments flag isn't set). And the link takes you to an empty create page instead of the list.

FredrikOseberg:
Skjermbilde 2022-03-21 kl  16 27 31

Should be a connected entity like before, with the stepper flowing "halfway" into the body of the info. But keep it as a single unit.

FredrikOseberg:
Skjermbilde 2022-03-21 kl  16 33 47

Note from UX team: 
* Stepper should have slightly different spacing
* Stepper circles should be adjusted according to sketch and use color scheme from sketch

src/component/segments/EditSegment/EditSegment.tsx
```diff
@@ -0,0 +1,99 @@
...
+    };
+
+    return (
+        <FormTemplate
+            loading={loading}
+            title="Create segment"
+            description="Segment makes it easy for you to define who should be exposed to your feature.
+            The segment is often a collection of constraints and can be reused.
+            Create a segment and use it when you configure an activation strategy on a feature toggle. "
+            documentationLink="https://docs.getunleash.io/how-to/how-to-create-and-assign-custom-project-roles"
```
thomasheartman, line 75: @FredrikOseberg 

I know this is beyond the scope of this PR, but I think we should update the form template component to make the "Learn more" link say something more meaningful:

https://www.a11yproject.com/posts/creating-valid-and-accessible-links/#write-helpful-link-text

A simple solution would be to add "about X" after "Learn more". If it's an issue that the link text gets longer and more unpredictable, we could also hide the extra text from rendering by using a `visually-hidden` class or something, but it'd be nice to have it there.

src/component/segments/CreateSegment/CreateSegment.tsx
```diff
@@ -0,0 +1,99 @@
...
+    };
+
+    return (
+        <FormTemplate
+            loading={loading}
+            title="Create segment"
+            description="Segment makes it easy for you to define who should be exposed to your feature.
+            The segment is often a collection of constraints and can be reused.
+            Create a segment and use it when you configure an activation strategy on a feature toggle. "
+            documentationLink="https://docs.getunleash.io/how-to/how-to-create-and-assign-custom-project-roles"
```
thomasheartman, line 75: I'm sure you're aware of this, but the link isn't correct at the moment. We should make sure to update this before it gets merged. Of course, we'll need to write the docs first 😅

src/component/segments/CreateSegment/CreateSegment.tsx
```diff
@@ -0,0 +1,99 @@
...
+        }
+    };
+    const handleCancel = () => {
+        history.goBack();
+    };
+
+    return (
+        <FormTemplate
+            loading={loading}
+            title="Create segment"
```
thomasheartman, line 71: The two calls to create and edit template seem awfully similar. If this happens for every feature, is there a way to abstract it in such a way that you don't need to provide the description and links twice? 🤔  Might not be, of course; just seems that it might be easier.

src/component/segments/CreateSegment/CreateSegment.tsx
```diff
@@ -0,0 +1,99 @@
...
+            description="Segment makes it easy for you to define who should be exposed to your feature.
+            The segment is often a collection of constraints and can be reused.
+            Create a segment and use it when you configure an activation strategy on a feature toggle. "
```
thomasheartman, line 74: Nice! 😄 I suggest a little rewrite into something like this (prettify as you wish):

```suggestion
            description="Segments make it easy for you to define which of your users should get access to a feature.
            A segment is a reusable collection of constraints.
            You can create and apply a segment when configuring activation strategies for a feature toggle or at any time from the segments page in the navigation menu."
```

thomasheartman (changes requested):
Nice work! 🎉  I've only had a look at the things Fredrik suggested (and little bits around it). Comments, questions, and suggestions are inline. Keep up the good work 🙌🏼

src/component/segments/EditSegment/EditSegment.tsx
olav: Good idea! I'll add an optional custom docs link label and use it in the segment forms, at least.

src/component/common/ConstraintAccordion/ConstraintAccordionList/ConstraintAccordionList.tsx
```diff
@@ -0,0 +1,142 @@
...
+export interface IConstraintAccordionListRef {
+    addConstraint: (contextName: string) => void;
+}
+
+// Extra form state for each constraint.
+interface IConstraintAccordionListItemState {
+    // Is the constraint currently being edited?
+    editing?: boolean;
+    // Is the constraint new (not yet saved)?
+    unsaved?: boolean;
```
thomasheartman, line 35: Is this true only if the constraint is new or is it also true if the constraint has unsaved changes?

src/component/common/FormTemplate/FormTemplate.tsx
```diff
@@ -144,12 +146,14 @@ const MobileGuidance = ({
 interface IGuidanceProps {
     description: string;
     documentationLink: string;
+    documentationLinkLabel?: string;
```
thomasheartman, line 149: Can we make the link label a required props "Learn more" is never a descriptive link label, so we should always provide more information. It would require changes in a lot of unrelated components, however, so it's probably more at home in a separate PR, but I'd love to hear some thoughts on this.

thomasheartman (approved):
All the changes I requested have been addressed and none of my new comments are blockers, so I can approve my part 🥳  However, I can't vouch for the rest of the code (too much context I don't have), so I'll leave that to someone else for now.

There's a few inline comments: one's a question (mostly about curiosity, but it means that I didn't understand the code/naming well enough) and one's a request / discussion point that I'd like to hear some thoughts around 😇

src/component/common/ConstraintAccordion/ConstraintAccordionList/ConstraintAccordionList.tsx
olav: Right, this is only `true` when the constraint is brand new (i.e. it has never been saved), used by the cancel button.

src/component/common/FormTemplate/FormTemplate.tsx
olav: Yeah, that would be great. A separate PR makes sense to me as well.

src/component/common/ConstraintAccordion/ConstraintAccordionList/ConstraintAccordionList.tsx
thomasheartman: Got it! I don't know if it needs a better name or not, it just wasn't clear to me when looking at the diff. Maybe `new`/`isNew` would be better? But again, if it's clear from looking at the code, then that's probably enough. An outside view isn't always helpful for things like this 💁🏼

olav: Definitely, `new` seems much better. Will fix.

src/component/common/FormTemplate/FormTemplate.tsx
thomasheartman: Sweet; glad you agree 😄 I went ahead and made a [Trello ticket for it](https://trello.com/c/8qX1K0nC/524-documentation-links-are-not-descriptive)

src/component/common/ConstraintAccordion/ConstraintAccordionList/ConstraintAccordionList.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+        const onCancel = (index: number) => {
+            const constraint = constraints[index];
+            state.get(constraint)?.new && onRemove(index);
+            state.set(constraint, {});
+        };
+
+        const onRemove = (index: number) => {
+            const constraint = constraints[index];
+            state.set(constraint, {});
```
FredrikOseberg, line 85: If I'm understanding this correctly, this WeakMap is a key value store where the object is used as the key and the metadata is associated with the object reference? Does it trigger re-renders?

I'm curious about this pattern.

src/component/common/FormTemplate/FormTemplate.styles.ts
```diff
@@ -1,5 +1,7 @@
 import { makeStyles } from '@material-ui/core/styles';
 
+export const formTemplateSidebarWidth = '27.5rem';
```
FredrikOseberg, line 3: Should this make it into the theme? There's a discussion regarding how to use the theme properly that I'd like to have at some point.

src/component/feature/FeatureStrategy/FeatureStrategyForm/FeatureStrategyForm.tsx
```diff
@@ -88,39 +82,44 @@ export const FeatureStrategyForm = ({
...
+                condition={Boolean(uiConfig.flags.SE)}
+                show={
+                    <FeatureStrategySegment
+                        segments={segments}
+                        setSegments={setSegments}
+                    />
+                }
+            />
             <ConditionallyRender
-                condition={Boolean(uiConfig.flags[C])}
```
FredrikOseberg, line 112: How is this working on main when flags is an object?

src/component/feature/FeatureStrategy/FeatureStrategySegment/FeatureStrategySegmentList.tsx
```diff
@@ -0,0 +1,54 @@
...
+                <Fragment key={segment.id}>
+                    <span className={styles.segment}>
+                        <Chip
+                            label={segment.name}
+                            onClick={onOpen(segment.id)}
+                            onDelete={onRemove(segment.id)}
+                        />
+                    </span>
+                    <ConditionallyRender
+                        condition={i < segments.length - 1}
```
FredrikOseberg, line 47: Could extract this in a variable `lastSegment`, I know this is the expression for the last element of an array - yet I still had to stop and think through what is going on here. Small readability optimisation.

src/component/segments/SegmentFormStepList/SegmentFormStepList.tsx
```diff
@@ -0,0 +1,35 @@
...
+                </span>
+                {Array.from({ length: total }).map((v, i) => (
+                    <FiberManualRecord
+                        key={i}
+                        className={classNames(
+                            styles.circle,
+                            i === current - 1 && styles.filledCircle
+                        )}
+                    />
+                ))}
```
FredrikOseberg, line 31: Could we optimise readability here? 

```suggestion
                {Array.from({ length: total }).map((_, index) => {
                    const currentStep = index  === current - 1
                    return <FiberManualRecord
                        key={i}
                        className={classNames(
                            styles.circle,
                            currentStep && styles.filledCircle
                        )}
                    />
                })}
```

src/component/segments/SegmentFormStepOne/SegmentFormStepOne.tsx
```diff
@@ -0,0 +1,83 @@
...
+    return (
+        <div className={styles.form}>
+            <h3 className={styles.formHeader}>Segment information</h3>
+            <div className={styles.container}>
+                <p className={styles.inputDescription}>
+                    What is the segment name?
+                </p>
+                <Input
+                    className={styles.input}
+                    label="Segment name*"
```
FredrikOseberg, line 40: Could be solved with required prop?

src/component/segments/SegmentFormStepTwo/SegmentFormStepTwo.tsx
```diff
@@ -0,0 +1,135 @@
...
+                <div className={styles.addContextContainer}>
+                    <p className={styles.inputDescription}>
+                        Or create and add a new custom context field
+                    </p>
+                    <SidebarModal
+                        label="Create new context"
+                        onClose={() => setOpen(false)}
+                        open={open}
+                    >
+                        <CreateContext
```
FredrikOseberg, line 76: Context is such a loaded word. I frequently mix it with react context, golang contexts and unleash context. To add to the confusion, there are subtle differences conceptually. Maybe we should rename our internal context references to `CreateUnleashContext` like we have done with the hook?

src/component/segments/SegmentList/SegmentList.tsx
```diff
@@ -109,73 +109,73 @@ export const SegmentsList = () => {
...
+                nextPage={nextPage}
+                prevPage={prevPage}
+                style={{ position: 'static', marginTop: '2rem' }}
+            />
+
+            <ConditionallyRender
+                condition={segments.length === 0}
+                show={renderNoSegments()}
+            />
+            {currentSegment && (
```
FredrikOseberg, line 169: Could also use ConditionallyRender here

FredrikOseberg (commented):
Looks really good so far! I'll take another look when it's completely ready.

src/component/common/ConstraintAccordion/ConstraintAccordionList/ConstraintAccordionList.tsx
olav: Yeah, just a way to associate extra fields with an object, without extending the object itself. The `useWeakMap` hook takes care of triggering re-renders whenever the map updates.

src/component/common/FormTemplate/FormTemplate.styles.ts
olav: Not sure, quite hacky at the moment. Only used by the form steps container, to center it above the main form column.

PR #787, opened 2022-03-15
refactor: improve strategy card focus styles

olav:


src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCard/FeatureStrategyMenuCard.styles.ts
```diff
@@ -12,7 +12,7 @@ export const useStyles = makeStyles(theme => ({
         borderStyle: 'solid',
         borderColor: theme.palette.grey[400],
         borderRadius: theme.spacing(1),
-        '&:hover': {
+        '&:hover, &:focus': {
```
thomasheartman, line 15: Nice! Just wondering, should we use `:focus` or `:focus-visible` here? Do we want the links to also get the focus styles when clicked?

thomasheartman (approved):
Looks good to me! Just an inline question we should think about, but it's hardly a blocker 😄

src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenuCard/FeatureStrategyMenuCard.styles.ts
olav: Good question. I don't think it matters here since we still show the default outline. The browser will still use its internal focus-visible rules for the outline, as far as I can tell.

PR #788, opened 2022-03-15
fix: make sure we handle hours as 00-23

chriswk:
So we were using date-fn's format library with `kk` which handles hours as 01-24, instead of HH which is 00-23.


chriswk:
[Constraint - date editor resets when selecting 00 hour](https://trello.com/c/OkNYBvft/954-constraint-date-editor-resets-when-selecting-00-hour)

sighphyre (approved):
Awesome!

Do we need a test for this so we catch regressions?

ykhedher (approved):

chriswk:
Yeah, we should've had a test in place here already. Let me see what I can do.

PR #794, opened 2022-03-16
refactor: fix search crash on missing values

olav:


src/hooks/useFeaturesFilter.ts
```diff
@@ -111,7 +111,6 @@ const filterFeatureByRegExp = (
     return feature.strategies.some(
         s =>
```
FredrikOseberg, line 112:
```suggestion
        strategy =>
```

src/hooks/useFeaturesFilter.ts
```diff
@@ -111,7 +111,6 @@ const filterFeatureByRegExp = (
...
             regExp.test(s.name) ||
-            // @ts-expect-error
-            s.constraints.some(c => c.values.some(v => regExp.test(v)))
+            s.constraints.some(c => c.values?.some(v => regExp.test(v)))
```
FredrikOseberg, line 114:
```suggestion
            strategy.constraints.some(constraint => constraint.values?.some(value => regExp.test(value)))
```

FredrikOseberg (changes requested):
Compilator will handle this for us, I need to stay sane :)

FredrikOseberg (approved):
LGTM

PR #805, opened 2022-03-22
fix: change rollout datatype to int

sighphyre:
GradualRollout has the rollout property specified as a string in both the client spec tests and our API docs and the Rust SDK expects that contract to be upheld. This commit changes the datatype of rollout from an int to a string


chriswk (approved):
Isn't this changing rollout datatype to string?

All your tests seem to point towards it  now becoming a string, except the `parseInt` function call?

sighphyre:
> Isn't this changing rollout datatype to string?
> 
> All your tests seem to point towards it now becoming a string, except the `parseInt` function call?

Yes, yes it is. This will teach me to write commit messages at the tail end of the day when I'm tired

FredrikOseberg:
> > Isn't this changing rollout datatype to string?
> > All your tests seem to point towards it now becoming a string, except the `parseInt` function call?
> 
> Yes, yes it is. This will teach me to write commit messages at the tail end of the day when I'm tired

We should add validation to the backend aswell.

PR #809, opened 2022-03-23
refactor: restrict API tokens to enabled environments

olav:


ykhedher (approved):
Nice work 🎉 .
I think we need to also block it from the backend as we can create a token for a disabled env from the API command.

PR #812, opened 2022-03-24
refactor: change strategy form buttons

ykhedher:


ykhedher:
[update strategy buttons form to use Create and Update button](https://trello.com/c/fK7Yyo3A/979-update-strategy-buttons-form-to-use-create-and-update-button)

thomasheartman (commented):
Is this related to this screenshot that I sent you a while back? 

image

With the new toggle overview, is that button still there? I've been trying to look for the changes, but I'm not sure where they are in the UI 🙈

ykhedher:
> Is this related to this screenshot that I sent you a while back?
> 
> image
> 
> With the new toggle overview, is that button still there? I've been trying to look for the changes, but I'm not sure where they are in the UI 🙈

Yes but I think we did remove that button from the UI
Screen Shot 2022-03-24 at 9 22 23 AM

thomasheartman:
In that case, shouldn't we just delete this button (or file)?

ykhedher:
> In that case, shouldn't we just delete this button (or file)?

Oh, I think I misunderstood you @thomasheartman.
This is the buttons in the create custom strategy. They are not using the CreateButton and UpdateButton components.

thomasheartman:
Ah, okay, so this is just replacing the custom strategy buttons with the predefined components? I think I get it now. If I read it right, there's also no visual change, right?

ykhedher:
Yes. Just following the same pattern

FredrikOseberg (approved):
LGTM

PR #820, opened 2022-03-25
chore: update paths 🧼

ykhedher:


ykhedher:
[clean front-end import paths 🧼](https://trello.com/c/O4JoorYm/985-clean-front-end-import-paths-%F0%9F%A7%BC)

sighphyre (approved):
This looks good to me. I think we should review the front end tests though, and make sure there's a test that touches each file at some point. It would make simple but sweeping changes like this much safer. This does seem to have broken some of the Cypress tests though, can we take a look into that before we merge?

olav (approved):

PR #826, opened 2022-03-29
feat: add FeedbackCES component

olav:
https://trello.com/c/DbLwMSw3/988-frontend-create-customer-effort-score-ces-feedback-component


src/app.css
```diff
@@ -150,3 +150,12 @@ p {
...
     }
 }
+
+.screen-reader-only {
+    position: absolute;
+    left: -10000px;
+    top: auto;
+    width: 1px;
+    height: 1px;
+    overflow: hidden;
```
thomasheartman, line 160: Nice! I'm not a hundred percent sure about this, but I think there is a 'more recent' way to do this, which covers some issues the 'position: absolute' method uses. It's covered in [this CSS-tricks post](https://css-tricks.com/inclusively-hidden/) (taken from [Inclusively Hidden](https://www.scottohara.me/blog/2017/04/14/inclusively-hidden.html) by Scott O'Hara)) among other things. 

```css
.screen-reader-only:not(:focus):not(:active) {
  clip: rect(0 0 0 0); 
  clip-path: inset(50%);
  height: 1px;
  overflow: hidden;
  position: absolute;
  white-space: nowrap; 
  width: 1px;
}
```

I've personally used a variant of that, which looks like this

```css
    border: 0;
    clip: rect(0 0 0 0);
    height: auto;
    margin: 0;
    overflow: hidden;
    padding: 0;
    position: absolute;
    width: 1px;
    white-space: nowrap;
```

Scott's equivalent would be (and I'd probably recommend you use this one over mine): 

```suggestion
    clip: rect(0 0 0 0); 
    clip-path: inset(50%);
    height: 1px;
    overflow: hidden;
    position: absolute;
    white-space: nowrap; 
    width: 1px;
```

I'll let you decide what the right way forward here is (you may know more about this than me!), but I'd generally trust Chis Coyier and Scott O'Hara on these things. The `:not(:focus):not(:active)` seems to be to allow the text to become visible when the element is focused (like with skip links).

src/component/feedback/FeedbackCES/FeedbackCES.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    const modalContent = state && (
+        <FeedbackCESForm state={state} onClose={hideFeedbackCES} />
+    );
+
+    return (
+        <Modal
+            open={Boolean(state)}
+            onClose={hideFeedbackCES}
+            disableEscapeKeyDown
```
thomasheartman, line 33: I'm not overly familiar with the Material Modal API, but does this disable closing on pressing `esc`? Is there any reason we're doing that? Can the user close the modal otherwise without submitting feedback?

There seems to be a close button, so `esc` should close the modal, as specified in the [WAI-aria authoring practices](https://w3c.github.io/aria-practices/#keyboard-interaction-7).

src/component/feedback/FeedbackCES/sendFeedbackInput.ts
```diff
@@ -0,0 +1,40 @@
+import { IFeedbackCESForm } from 'component/feedback/FeedbackCES/FeedbackCESForm';
+
+interface IFeedbackEndpointRequestBody {
+    source?: 'docs' | 'app' | 'app:segments';
```
thomasheartman, line 4: The app should never send `docs` feedback, should it? I suggest removing `docs` from the type definition in the app, so as to avoid possible misfires 💁🏼

src/component/feedback/FeedbackCES/FeedbackCESScore.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface IFeedbackCESScoreProps {
+    form: Partial<IFeedbackCESForm>;
+    setForm: React.Dispatch<React.SetStateAction<Partial<IFeedbackCESForm>>>;
+}
+
+export const FeedbackCESScore = ({ form, setForm }: IFeedbackCESScoreProps) => {
+    const styles = useStyles();
+
+    const onScoreChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        event.persist();
```
thomasheartman, line 15: What version of React are we using? [According to the docs, `event.persist` doesn't do anything from React 17 an onwards](https://reactjs.org/docs/legacy-event-pooling.html), so I'm not sure this is necessary.

src/component/feedback/FeedbackCESContext/FeedbackCESProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+    // Set a new feedback state iff the path is unseen and CES is enabled.
+    const showFeedbackCES: ShowFeedbackCES = useCallback(
+        value => {
+            setState(prev => {
+                const next = value instanceof Function ? value(prev) : value;
+                return !next || seen[next.path] || !enabled ? undefined : next;
+            });
+        },
+        [enabled, seen]
+    );
```
thomasheartman, line 46: For clarification: is everything done via localstorage? Would that be an issue for users? Are they gonna switch browsers often? I think the other in-app feedback uses a dedicated db table. Is that something we want to consider here too?

thomasheartman (changes requested):
Nice one 🙌🏼  I've got a few questions and comments inline. It's mostly good, but some things require action before merging I think 😄

src/component/feedback/FeedbackCES/sendFeedbackInput.ts
thomasheartman: And in doing so, I'd also make `source` a required property.

src/component/feedback/FeedbackCES/FeedbackCESScore.tsx
thomasheartman: I mean, I suppose I could check the package.json file, but why spend 30 seconds doing that when you can spend way longer reading about details 😅  But yeah, the package.json says v17.0.2, so I don't think we need this anymore 😄 

image

src/app.css
thomasheartman: More content:

[WebAIM seems to say absolute positioning and clipping do the same thing](https://webaim.org/techniques/css/invisiblecontent/) and doesn't recommend one over the other. The article was last updated in September 2020, so it's probably fine 🤓

src/component/feedback/FeedbackCES/FeedbackCESScore.tsx
olav: Nice catch, I thought that was a v18 thing!

src/app.css
olav: Great, thanks, seems much better. 

We don't really need a generic solution here, so I've moved the required styles to where they're used.

src/component/feedback/FeedbackCES/FeedbackCES.tsx
olav: Yeah, agreed. There was a specific request for `esc` to be disabled, but will remove this for now.

src/component/feedback/FeedbackCESContext/FeedbackCESProvider.tsx
olav: That's much better! The seen state should now be stored on the backend.

thomasheartman (commented):
The code looks good to me but I have a few questions which you can choose to act on or not:

1. Can/should we test this is in some way? I don't know how we usually test the UI, if at all. Is it worth doing for this?

2. Can you confirm that when the modal is active:
  - clicking outside the modal closes the modal
  - when the modal is active, it traps focus
  - when the modal is closed, focus is returned to where it was before it opened (if that's the right behovior in this case)
  
I'm assuming that MUI takes care of point 2, but you never really know 💁🏼

olav:
> 1. Can/should we test this is in some way? I don't know how we usually test the UI, if at all. Is it worth doing for this?

Good idea. I've added a little snapshot test at least. We can spend more time on better tests if the feature turns brittle.

> 2. Can you confirm that when the modal is active: [...]

Yep, these seem to work as expected.

thomasheartman (approved):
Thanks for all the hard work and for listening to my feedback! 😄 

For my part, at least, I'd say this is ready to go 🚂

PR #831, opened 2022-03-30
fix: toast in mobile view

ykhedher:


ykhedher:
[Mobile - Improvements](https://trello.com/c/G2mSkULI/998-mobile-improvements)

src/common.styles.ts
```diff
@@ -24,12 +24,14 @@ export const useCommonStyles = makeStyles(theme => ({
     },
     divider: {
         margin: '1rem 0',
+        //@ts-expect-error
```
olav, line 27: Would add s space here, like `// @ts-expect-error`.

src/common.styles.ts
```diff
@@ -24,12 +24,14 @@ export const useCommonStyles = makeStyles(theme => ({
...
         backgroundColor: theme.palette.division.main,
         height: '3px',
         width: '100%',
     },
     largeDivider: {
         margin: '2rem 0',
+        //@ts-expect-error
```
olav, line 34: Same here.

src/common.styles.ts
```diff
@@ -112,6 +114,11 @@ export const useCommonStyles = makeStyles(theme => ({
         right: '40px',
         top: '40px',
         transform: 'translateY(-400px)',
+        [theme.breakpoints.down('sm')]: {
+            right: 20,
+            left: 10,
```
olav, line 119: Should the feedback box stay top right on a tablet? Maybe it's enough to remove `left` here. :thinking:

olav (changes requested):

src/common.styles.ts
ykhedher: @olav If I remove the left the feedback dialog will have no space from the left. It will look like this:
image

and this is how it looks like if I keep the left:
image

olav (approved):

PR #837, opened 2022-03-31
feat: show segments in feature overview

olav:


sighphyre (approved):
LGTM but you might want to double check with someone with better front end eyes than myself

PR #843, opened 2022-04-01
refactor: port some things to TS

olav:


src/component/tags/TagTypeList/TagTypeList.tsx
```diff
@@ -123,10 +129,11 @@ export const TagTypeList = () => {
                 <ListItemText primary={link} secondary={tagType.description} />
                 <PermissionIconButton
                     permission={UPDATE_TAG_TYPE}
+                    // @ts-expect-error
```
nunogois, line 132: Can you please expand a bit on why you need this here?

nunogois (approved):
Overall it looks pretty good to me :+1:

src/component/tags/TagTypeList/TagTypeList.tsx
olav: Nice catch! Just a pending TODO, but why not fix the `PermissionIconButton` props right away. Should be better now.

PR #847, opened 2022-04-04
use relative path for api

JuliaHasCats:
Our company has unleash running behind a reverse proxy server, so unleash api endpoints are at https://mydomain.com/unleash/api/*
With the formatApiPath function returning absolute path, the front end is trying to reach unleash apis at https://mydomain/api/*, casuing front end unable to load as it cannot reach unleash apis.

Api path formatting should follow the same way as formatting asset path.

### DIFF
--- a/src/component/providers/SWRProvider/SWRProvider.tsx
+++ b/src/component/providers/SWRProvider/SWRProvider.tsx
@@ -30,10 +30,10 @@ const SWRProvider: React.FC<ISWRProviderProps> = ({
             }
 
             if (
-                path === formatApiPath('login') ||
-                path === formatApiPath('new-user') ||
-                path === formatApiPath('reset-password') ||
-                path === formatApiPath('forgotten-password')
+                path === formatApiPath('/login') ||
+                path === formatApiPath('/new-user') ||
+                path === formatApiPath('/reset-password') ||
+                path === formatApiPath('/forgotten-password')
             ) {
                 return;
             }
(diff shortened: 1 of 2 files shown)


ivarconr:
@JuliaHasCats This should not be necessary, have you tried the `baseUriPath` option in Unleash ([docs](https://docs.getunleash.io/deploy/configuring_unleash#further-customization))?

FredrikOseberg:
> 

@JuliaHasCats 

Not sure I understand the issue here. If you have unleash running on subpath like: `https://mydomain.com/unleashpath` and you set the configuration of the unleash instance to be: 

```
 server: {
              baseUriPath: 'unleashpath',
 },
```

This will be encoded into the html file and retrieved by the frontend to build the correct path. What am I missing regarding your use case?

JuliaHasCats:
> > 
> 
> @JuliaHasCats
> 
> Not sure I understand the issue here. If you have unleash running on subpath like: `https://mydomain.com/unleashpath` and you set the configuration of the unleash instance to be:
> 
> ```
>  server: {
>       baseUriPath: 'unleashpath',
>  },
> ```
> 
> This will be encoded into the html file and retrieved by the frontend to build the correct path. What am I missing regarding your use case?

Thanks for replying. Yes, it would work for frontend, but this path is also going to be encoded into api endpoints. And with our weird situation, we cannot have APIs to be added a subpath.
We found a workaround for it. I'll just close this PR.

PR #851, opened 2022-04-05
refactor: add segment limit warnings

olav:


src/component/segments/SegmentDocs/SegmentDocs.tsx
```diff
@@ -0,0 +1,89 @@
...
+export const SegmentDocsWarning = () => {
+    const styles = useStyles();
+
+    return (
+        <Alert severity="warning">
+            <p className={styles.paragraph}>
+                Segments is an experimental feature available to select users.
+            </p>
+            <p className={styles.paragraph}>
+                This feature is currently in development and future versions may
```
thomasheartman, line 17: Minor phrasing suggestion. I feel like this would read better 💁🏼

```suggestion
                This feature is currently in development. Future versions may
```

src/component/common/AutocompleteBox/AutocompleteBox.tsx
```diff
@@ -41,6 +50,7 @@ export const AutocompleteBox = ({
                 onChange={(event, value) => onChange(value || [])}
                 renderInput={renderInput}
                 getOptionLabel={value => value.label}
+                disabled={disabled}
```
thomasheartman, line 53: This is outside my a11y chops, but I think it's something we should look into:

What happens when this is disabled? Does it disappear entirely for screen readers? Consider a case where a user uses this input to add multiple items. When it's full, does the input just disappear for the user? How would they know what's happened? 

When you disable an input, it becomes unfocusable (and thus unreachable) for users navigating by keyboards. I imagine (think I've read it somewhere) that this would also apply to screen readers. We might want to add an [aria-live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live) or similar to alert users that the input has become disabled and that they can't add any more items.

What do you think?

thomasheartman (changes requested):
Looks mostly good to me 🙌🏼  I've got one suggestion regarding wording and a potential a11y issue that I'd love to have addressed. I'm not sure it _is_ an issue, however, but I'd like to just discuss it before approving ☺️

src/component/segments/SegmentDocs/SegmentDocs.tsx
olav: True, much better!

src/component/common/AutocompleteBox/AutocompleteBox.tsx
olav: Good question! Tried it out using VoiceOver on macOS: 

When adding segments to strategies, and the limit gets reached, we show the warning box. This warning box gets announced due to its `role="alert"` attribute. The `Autocomplete` is still focusable, just announced as "dimmed". 

Thereafter, the `Autocomplete` gets re-announced sometimes, but not always. Maybe some internal heuristic?

thomasheartman: Understood! That sounds like a good solution to me. Again, I'm no expert, but that sounds like a sensible way to handle it. 

I'd like to give this a go myself: how would I do that? Go to segments and add them until they're full?

thomasheartman (approved):

src/component/common/AutocompleteBox/AutocompleteBox.tsx
olav: Yeah, a start at least. There's probably lots more we could do here. The accessible description for the Autocomplete seems a bit messy. Maybe an upgrade to MUI v5 would help.

> I'd like to give this a go myself: how would I do that? Go to segments and add them until they're full?

Yeah! I think the easiest way would be to add 5 segments to a strategy. That should disable the box and show the warning.

thomasheartman: Cool! Where do I find segments? Are they available in the preview?

olav: For now I'm afraid you'll have to run everything locally to test the latest stuff. But should be out in multiple envs pretty soon.

PR #853, opened 2022-04-05
refactor: use --frozen-lockfile in node action

olav:


.github/workflows/release.yml
```diff
@@ -38,4 +38,3 @@ jobs:
     - name: Publish static assets to S3
       run: |
         aws s3 cp build/ s3://getunleash-static/unleash/${{ steps.get_version.outputs.VERSION }} --recursive
```
Tymek, line 40: `.editorconfig` has `insert_final_newline = true`. weird that it is removed here

Tymek (approved):

.github/workflows/release.yml
olav: Looks like there were two trailing newlines in main.

PR #872, opened 2022-04-08
fix: remove separator from first option in list

olav:
Fixes a case where we'd show a separator line above the first operator dropdown option.


nunogois (approved):
Looks good to me 👍

PR #874, opened 2022-04-11
feat: add context value descriptions

olav:


src/component/context/ContectFormChip/ContextFormChip.tsx
```diff
@@ -0,0 +1,30 @@
...
+    description,
+    onRemove,
+}: IContextFormChipProps) => {
+    const styles = useStyles();
+
+    return (
+        <li className={styles.container}>
+            <div>
+                <div className={styles.label}>{label}</div>
+                {description && (
```
FredrikOseberg, line 21: I would prefer `ConditionallyRender` here, even though this is a small check. The `&&` pattern can be dangerous (not with this input, but in other contexts - https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx) 

If we don't use it in the codebase it will discourage new developers from using this pattern.

FredrikOseberg (changes requested):
There seems to be some issues with the legal value input.

Problem 1:
If I use `Enter` to add a value, it will clear the input fields and give no feedback. Previously it would add the legal value. 

Problem 2:
If I add a legal value: `my legal value` and then try to add the same legal value by hitting enter it will give me an error message, but remove the legal value from the rendered output and sidebar payload.

Problem 3:
Writing in the input or blurring it does not clear the error message

Video summary: https://www.loom.com/share/3b1b99f4a49c4eaab9f1a638e81a0cac

Seems to be connected the enter event.

NicolaeUnleash (changes requested):
1. We need to reduce the space between the inputs like in the image to 0.5rem
image

2. We need a validation if you want to add a a value only with description:
image

src/component/context/ContextForm/ContextForm.tsx
```diff
@@ -45,54 +47,64 @@ export const ContextForm: React.FC<IContextForm> = ({
...
     const [value, setValue] = useState('');
-    const [focused, setFocused] = useState(false);
+    const [valueDesc, setValueDesc] = useState('');
+    const [valueFocused, setValueFocused] = useState(false);
 
-    const submit = (event: React.SyntheticEvent) => {
+    const isMissingValue = valueDesc.trim() && !value.trim();
+
+    const isDuplicateValue = legalValues.some(legalValue => {
+        return legalValue.value.trim() === value.trim();
```
FredrikOseberg, line 56: This causes a crash when opening existing context fields without legal values:

Skjermbilde 2022-04-19 kl  11 12 09

FredrikOseberg (changes requested):
Close, but a bit nasty crash on existing context fields we need to fix. 

This causes the page content to jump when an error message is displayed. We have the Input component that will display the error message absolutely relative to the parent input and avoid this jumping, but it might require more space between the input fields. WDYT @NicolaeUnleash and @olav  ?

Skjermbilde 2022-04-19 kl  11 19 04

olav:
> This causes the page content to jump when an error message is displayed.

Yeah, would be nice to avoid the jump! But we'd have to increase the spacing quite a bit. Could perhaps move the error message below both fields, though that seems less accessible.

FredrikOseberg (approved):
LGTM. If we can smoothe out the transition when jump occurs, that would be great. Otherwise, we'll live with it.

PR #877, opened 2022-04-13
fix: make sure the override context name exists

olav:


FredrikOseberg (approved):
LGTM. Might need to consider the corner-case if all context fields are deleted in the future - but that might be another PR.

PR #879, opened 2022-04-13
feat: add user table sort and search

nunogois:
https://trello.com/c/iamqRWiU


src/component/admin/users/UsersList/UsersList.tsx
```diff
@@ -49,8 +57,14 @@ const UsersList = () => {
     const [inviteLink, setInviteLink] = useState('');
     const [delUser, setDelUser] = useState<IUser>();
     const ref = useLoading(loading);
+    const { filtered, setFilter } = useUsersFilter(users);
+    const { sorted, sort, setSort } = useUsersSort(filtered);
     const { page, pages, nextPage, prevPage, setPageIndex, pageIndex } =
-        usePagination(users, 50);
+        usePagination(sorted, 50);
```
FredrikOseberg, line 63: I believe what you have done here should work, but can you verify that reducing the number of users per page and sorting also affects the pagination UI?

src/component/common/AnimateOnMount/AnimateOnMount.tsx
```diff
@@ -27,6 +31,7 @@ const AnimateOnMount: FC<IAnimateOnMountProps> = ({
         if (mountedRef.current !== mounted || mountedRef === null) {
             if (mounted) {
                 setShow(true);
+                onStart?.();
```
FredrikOseberg, line 34: Is this a special syntax for using the ? operator with a function? It looks to me like it's trying to access a property on the function.

src/component/common/Table/TableActions/TableActions.styles.ts
```diff
@@ -0,0 +1,65 @@
...
+            width: 0,
+            transition: 'all 0.6s',
+        },
+        '& .search-container': {
+            borderColor: 'transparent',
+        },
+    },
+    verticalSeparator: {
+        height: '100%',
+        backgroundColor: '#BDBDBF',
```
FredrikOseberg, line 57: This is also something that should be codified into the theme.

src/component/common/Table/TableActions/TableActions.styles.ts
```diff
@@ -0,0 +1,65 @@
...
+        width: '45px',
+        marginRight: 0,
+        transition: 'all 0.6s',
+        '& .search-icon': {
+            marginRight: 0,
+            transition: 'all 0.6s',
+        },
+        '& .input-container, .clear-container': {
+            width: 0,
+            transition: 'all 0.6s',
```
FredrikOseberg, line 49: We should be explicit with setting transitions. When we use _all_ we are setting up listeners to look for every property change. Instead set a transition on the property you want to change, i.e: `transition: translate 0.6s`

src/component/common/Table/TableActions/TableActions.styles.ts
```diff
@@ -0,0 +1,65 @@
...
+            borderColor: theme.palette.grey[300],
+        },
+    },
+    fieldWidthLeave: {
+        width: '45px',
+        marginRight: 0,
+        transition: 'all 0.6s',
+        '& .search-icon': {
+            marginRight: 0,
+            transition: 'all 0.6s',
```
FredrikOseberg, line 45: See comment below:

src/component/common/Table/TableActions/TableActions.tsx
```diff
@@ -0,0 +1,59 @@
...
+import AnimateOnMount from 'component/common/AnimateOnMount/AnimateOnMount';
+import { TableSearchField } from 'component/common/Table/TableActions/TableSearchField/TableSearchField';
+import { useStyles } from 'component/common/Table/TableActions/TableActions.styles';
+
+interface ITableActionsProps {
+    search: string;
+    onSearch: (value: string) => void;
+}
+
+export const TableActions = ({ search, onSearch }: ITableActionsProps) => {
```
FredrikOseberg, line 14: Nice abstraction

src/component/common/Table/TableActions/TableSearchField/TableSearchField.styles.ts
```diff
@@ -0,0 +1,39 @@
...
+    container: {
+        display: 'flex',
+        alignItems: 'center',
+        flexWrap: 'wrap',
+        gap: '1rem',
+    },
+    search: {
+        display: 'flex',
+        alignItems: 'center',
+        // @ts-expect-error
```
FredrikOseberg, line 13: This should be solved by augmenting the theme type instead of setting `// @ts-expect-error`

src/component/common/Table/TableCellSortable/TableCellSortable.styles.ts
```diff
@@ -0,0 +1,25 @@
+import { makeStyles } from '@material-ui/core/styles';
+
+export const useStyles = makeStyles(theme => ({
+    tableCellHeaderSortable: {
+        cursor: 'pointer',
+        '&:hover': {
+            backgroundColor: '#e6e6e9', // Should be a variable similar to theme.palette.grey[70] TODO: fix this color
```
FredrikOseberg, line 7: Agree. Should be a theme variable.

src/component/common/Table/TableCellSortable/TableCellSortable.styles.ts
```diff
@@ -0,0 +1,25 @@
...
+            '& > svg': {
+                color: '#000',
+            },
+        },
+        '& > svg': {
+            fontSize: '18px',
```
FredrikOseberg, line 13: This should also be a theme variable.

src/component/common/Table/TableCellSortable/TableCellSortable.styles.ts
```diff
@@ -0,0 +1,25 @@
+import { makeStyles } from '@material-ui/core/styles';
+
+export const useStyles = makeStyles(theme => ({
+    tableCellHeaderSortable: {
+        cursor: 'pointer',
+        '&:hover': {
+            backgroundColor: '#e6e6e9', // Should be a variable similar to theme.palette.grey[70] TODO: fix this color
+            '& > svg': {
+                color: '#000',
```
FredrikOseberg, line 9: And this.

src/hooks/useUsersFilter.ts
```diff
@@ -0,0 +1,62 @@
...
+const filterUsersByQuery = (users: IUser[], filter: IUsersFilter): IUser[] => {
+    if (!filter.query) {
+        return users;
+    }
+
+    // Try to parse the search query as a RegExp.
+    // Return all users if it can't be parsed.
+    try {
+        const regExp = new RegExp(filter.query, 'i');
+        return users.filter(f => filterUserByRegExp(f, regExp));
```
FredrikOseberg, line 50: What is _f_ in this context? It looks like a `user` to me? Let's change this to `users.filter(user => filterUserByRegExp(user, regExp))`

src/hooks/useUsersSort.ts
```diff
@@ -0,0 +1,114 @@
...
+): IUser[] => {
+    return [...users].sort((a, b) =>
+        getRoleName(a.rootRole, roles).localeCompare(
+            getRoleName(b.rootRole, roles)
+        )
+    );
+};
+
+const getRoleName = (roleId: number, roles: Readonly<IRole[]>) => {
+    const role = roles.find((r: IRole) => r.id === roleId);
```
FredrikOseberg, line 112: Let's use the full parameter name for role here 
```suggestion
    const role = roles.find((role: IRole) => role.id === roleId);
```

FredrikOseberg (changes requested):
In general I think this looks very nice. There are some smaller things to clean up and align on - such as theming - and I would like to see some tests for the sorting

package.json
```diff
@@ -76,6 +77,7 @@
     "react-dnd": "15.1.2",
     "react-dnd-html5-backend": "15.1.3",
     "react-dom": "17.0.2",
+    "react-highlight-words": "^0.18.0",
```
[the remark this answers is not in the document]
nunogois: Agreed, would prefer removing it or following a custom approach, as I'm guessing it wouldn't be too hard to create a component like this to suit our needs, but let's see what the consensus is.

src/component/common/AnimateOnMount/AnimateOnMount.tsx
nunogois: It's the optional chaining operator applied to a function: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Since onStart and onEnd are two optional props, this seems a bit cleaner than doing something like:
```JS
if (onStart) {
  onStart()
}
```

package.json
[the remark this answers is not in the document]
Tymek: Let's add [bundle analyzer](https://create-react-app.dev/docs/analyzing-the-bundle-size/).

src/hooks/useUsersFilter.ts
nunogois: You're right, I copied this from the existing `useFeaturesFilter` hook where `f` (as in feature) made more sense. Changed to `user` as per your suggestion 👍

src/component/admin/users/UsersList/UsersList.tsx
nunogois: Quickly tested with `usePagination(sorted, 2)` and everything seems to be behaving correctly, both for sorting and searching.

src/component/common/Table/TableActions/TableSearchField/TableSearchField.styles.ts
nunogois: Wasn't sure how to extend the theme to add this property, since all the current examples use `// @ts-expect-error`, so I made the change to `theme.palette.background.paper`, which maybe makes more sense, since we want this to be the color of the background paper, whatever that color is (and not something like `theme.palette.common.white` statically).

src/hooks/useUsersFilter.ts
```diff
@@ -0,0 +1,62 @@
...
+        if (err instanceof SyntaxError) {
+            return users;
+        } else {
+            throw err;
+        }
+    }
+};
+
+const filterUserByRegExp = (user: IUser, regExp: RegExp): boolean =>
+    regExp.test(user.name ?? '') ||
```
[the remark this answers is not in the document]
nunogois: @sonatype-lift ignore

package.json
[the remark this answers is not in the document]
nunogois: Created a lightweight `Highlighter` component based on our needs to address this question.

src/hooks/useUsersFilter.ts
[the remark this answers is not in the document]
nunogois: @sonatype-lift ignore

FredrikOseberg (approved):
LGTM

src/hooks/useUsersFilter.ts
```diff
@@ -0,0 +1,62 @@
...
+    regExp.test(user.username ?? user.email ?? '');
```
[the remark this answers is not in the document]
nunogois: @sonatype-lift ignore

src/hooks/useUsersFilter.ts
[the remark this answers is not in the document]
nunogois: @sonatype-lift ignore

PR #886, opened 2022-04-20
fix: improve HTML landmarks

olav:


thomasheartman:
@olav Maybe I'm missing something, but why are you changing all the headers to `h1`s? Is there some way to automatically use the right heading level based on that? As [the MDN docs for heading elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) mention, you should only have one `h1` on each page, so using them for footers etc seems a bit excessive 🤔

olav:
@thomasheartman You're right, those should be `<h2>`s. Should be fixed now.

Also fixed the header order on the feature overview page.

thomasheartman:
@olav Nice one! I'm assuming the rest of the h1s are page titles, then? 😄 

Did you look into adding a `nav` landmark, by the way (as mentioned in [the trello ticket](https://trello.com/c/Cft6Jn69/311-page-landmarks-add-main-section-consider-extra-nav-section))?

olav:
@thomasheartman Thanks for the link! Added `nav` and `search`. 

Also fixed a few more headings (there are probably more, but it's a start).

FredrikOseberg (approved):
LGTM

PR #887, opened 2022-04-20
fix: make environment switch text clickable

olav:


NicolaeUnleash (changes requested):
@olav we need to make the cursor pointer if we can click on the text

olav:
@NicolaeUnleash Should be fixed now!

NicolaeUnleash (approved):
looks good

PR #888, opened 2022-04-20
fix: slightly improve tabs markup

olav:
This PR fixes some invalid ARIA tab attributes. The next step could be to look at how the tabs mix buttons and links.


FredrikOseberg (approved):
LGTM

PR #891, opened 2022-04-20
refactor: remove default link color

olav:
Makes links match their parent by default.

image


nunogois (approved):

NicolaeUnleash:
this should be updated again when we align the color for the links in the entire app

PR #892, opened 2022-04-20
fix: improve avatar alt text

olav:
Avoid repeating text that's in the title attribute or right next to the image.


FredrikOseberg (approved):
LGTM

PR #898, opened 2022-04-22
refactor: use buttons for sortable <th>s

olav:


thomasheartman:
@olav Code looks good, but I'd like to test it out before I approve it. Can you tell me where I find this in the UI? I can't seem to find any tables right now 🙈

olav:
@thomasheartman Good idea! These are the sortable tables I found:

* https://unleash-frontend-git-table-sort-unleash-team.vercel.app/admin/users
* https://unleash-frontend-git-table-sort-unleash-team.vercel.app/projects/default

thomasheartman (approved):
Superb! ❤️ This is so much better than what it was 😄 

One thing (feel like I've raised this before): how does this work with screen readers (I still haven't set mine up 🙈)? Do they announce that the table gets reordered or should we add an aria-live region to it?

olav:
> One thing (feel like I've raised this before): how does this work with screen readers (I still haven't set mine up see_no_evil)? Do they announce that the table gets reordered or should we add an aria-live region to it?

Good point. Looks like one approach is to explicitly announce the new sort order through a live region. I've added a little Announcer component that we can use for such things.

thomasheartman:
Champion! 🥇

PR #900, opened 2022-04-22
fix: enable arrow key nav in header menus

olav:


src/component/menu/Header/NavigationLink/NavigationLink.styles.ts
```diff
@@ -23,6 +23,6 @@ export const useStyles = makeStyles(theme => ({
         color: '#000',
         height: '100%',
         width: '100%',
-        padding: '0.5rem 1rem',
+        padding: '0.5rem 1rem !important',
```
thomasheartman, line 26: Ooh, `!important`? Can we increase the specificity of this selector some other way to avoid this?

thomasheartman (approved):
Looks good to me 🥳  The only point of contention is the use of `!important` as called out inline. Can we increase the specificity of that selector some other way?

src/component/menu/Header/NavigationLink/NavigationLink.styles.ts
olav: Good idea. We can use the `classes` prop to override MUI styles instead.

olav: Good idea, we can use the `classes` prop to override MUI styles instead.

PR #913, opened 2022-04-26
refactor: remove form page autoFocus attributes

olav:


thomasheartman (commented):
Ooh, this may actually be a bit contentious. Why are you removing them? 

For a bit of context: I think _I_ was the one who added most of them because having to navigate through the whole page to get to the form input field after I've expressly asked to get a form is a bit of a pain. Still, it may not be the wrong thing, so I'd love to hear your thoughts!

olav:
@thomasheartman Oh good point! The `autofocus` attributes seemed confusing in a screen reader after adding the page title announcements, since the input field gets announced instead. 

Though reading up on it now, `autofocus` doesn't seem to be as problematic as I thought. We can keep them for now.

thomasheartman (commented):
Update: You got me real curious, so I had to go and do some reading on the topic 🤓 

It seems that it's ... contentious. It can easily be (and often is) abused. [The a11y project suggests to avoid using it](https://www.a11yproject.com/checklist/#global-code), but the [W3C author's guide it links to](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html) isn't so straightforward, saying that:

>The intent of this Success Criterion is to ensure that when users navigate sequentially through content, they encounter information in an order that is consistent with the meaning of the content and can be operated from the keyboard.

It also specifies that it's the Right Thing ™️  for modal forms that pop up, for instance. This, by the way, _was_ the case when I added the `autofocus` items. 

They're not modals anymore, but even now, the forms only exist on pages where they're the main reason for the page existing, so I tend to agree with a11y guru(?) Bruce Lawson when he says:

> I believe that autofocussing into a form is a usability win for most people if the purpose of the page is that form—for example, the Google hompage. So, I wouldn’t autofocus into the search box or comment field on this page because the primary purpose of this page is discussing this article which requires prior reading. Most people are lurkers, not commenters, so it’s against usability to autofocus into the comment form. But a contact form is the primary purpose of a contact page, and it saves users a keystroke to put the caret into the first field—particularly usable for mobile phone users.

In his article on [the accessibility of HTML 5 autofocus article](https://brucelawson.co.uk/2009/the-accessibility-of-html-5-autofocus/)

But, he _does_ go on to say that:

> However, screenreader users find themselves dumped in the middle of a form, potentially without context (athough it’s always possible to tell the screenreader to repeat the title of the page, and we all give our pages useful titles, don’t we?).
>
> Moreover, many forms have explanatory or introductory information preceeding them. My contact form does, although that’s somewhat flippant. A better example is [Salford University’s Online Staff Directory Quick Search Page](http://directory.salford.ac.uk/). The explanatory information is correctly outside the form element, as JAWS and (I think) Windows-Eye ignore any text that’s not inside legend and label elements when they’re in Forms mode.

He then goes on to suggest you use `aria-described-by` for the whole form if it's got descriptive content somewhere.

In all: I ... don't know. I'm leaning towards it making sense as long as the form is the primary content of the page (such as when logging in etc), but that maybe a (potentially invisible) form description might be useful.

Thoughts?

thomasheartman:
Also, ah! That makes sense with the page titles and how that works. I didn't realize that would be a thing. Hmm 🤔

olav:
Yeah definitely a tricky topic, something to think about. Thanks for the links! :eyes:

PR #914, opened 2022-04-26
refactor: misc login page accessibility improvements

olav:


thomasheartman (approved):
Looks good to me! But on a login page (going back to the PR): would it make sense to autofocus the first input field? It would make _my_ life a whole lot easier, at least, but if it's not The Way™️, then I'll respect that 💁🏼

olav:
@thomasheartman Good idea!

PR #923, opened 2022-04-29
chore: update MUI to v5

olav:


nunogois (approved):
I was not super thorough since there are a lot of changes, but overall it looks good to me. Great job 👍 

Also asked @NicolaeUnleash to have a look, just in case we have any weird or unexpected UI/UX with the upgrade that we should take note of.

PR #930, opened 2022-04-29
refactor: fix segment permission checks

olav:


sighphyre (approved):
Local testing confirms this fixes the issue!

PR #942, opened 2022-05-02
feat: show archived toggles on a project level

nunogois:
https://trello.com/c/lfxLejtK


src/hooks/api/getters/useFeaturesArchive/useFeaturesArchive.ts
```diff
@@ -9,10 +9,19 @@ export interface IUseFeaturesArchiveOutput {
     error?: Error;
 }
 
-export const useFeaturesArchive = (): IUseFeaturesArchiveOutput => {
+export const useFeaturesArchive = (
```
FredrikOseberg, line 12: I think I would prefer to separate this into a `useProjectArchive` hook instead of augmenting this hook with conditionals.

src/component/archive/ArchiveListContainer.tsx
```diff
@@ -6,7 +7,13 @@ import { useFeatureArchiveApi } from 'hooks/api/actions/useFeatureArchiveApi/use
 import useToast from 'hooks/useToast';
 import { useFeaturesSort } from 'hooks/useFeaturesSort';
 
-export const ArchiveListContainer = () => {
+interface IArchiveListContainer {
+    projectId?: string;
+}
+
+export const ArchiveListContainer: FC<IArchiveListContainer> = ({
```
FredrikOseberg, line 14: Instead of building this into the existing container, I think it would be better if we separated the two. Less to think about when using the components. So something like `<ProjectArchiveList />` that uses a dedicated hook for this. There might be some duplication of API logic, but less complexity and the components will only do one thing.

src/component/feature/FeatureToggleList/FeatureToggleListActions/FeatureToggleListActions.tsx
```diff
@@ -46,6 +48,10 @@ export const FeatureToggleListActions: VFC<IFeatureToggleListActionsProps> = ({
     const selectedOption =
         sortOptions.find(o => o.type === sort.type) || sortOptions[0];
 
+    if (inProject) {
+        sortOptions = sortOptions.filter(o => o.type !== 'project');
```
FredrikOseberg, line 52:
```suggestion
        sortOptions = sortOptions.filter(option => option.type !== 'project');
```

src/component/feature/FeatureToggleList/FeatureToggleListActions/FeatureToggleListActions.tsx
```diff
@@ -73,7 +79,7 @@ export const FeatureToggleListActions: VFC<IFeatureToggleListActionsProps> = ({
                 data-loading
             />
             <ConditionallyRender
-                condition={uiConfig.flags.P}
+                condition={uiConfig.flags.P && !inProject}
```
FredrikOseberg, line 82: I don't want to spend that much time on the actual FeatureToggleList component since we are going to make some changes here soon, but the more we can avoid having a component for two different use cases, the easier it will be to work with. We should strive to lift these things up into the parent components where possible and use composition to change elements that are more deeply nested.

FredrikOseberg (changes requested):
This is a good start. There are small improvements that can be done, but it's pretty close to done.

src/hooks/api/getters/useFeaturesArchive/useFeaturesArchive.ts
nunogois: Makes sense. Would `useProjectFeaturesArchive` be clearer? Who knows, we might have a project archive (archiving the projects themselves) in the future.

FredrikOseberg (approved):
Nice. LGTM!

PR #944, opened 2022-05-03
refactor: clean up strategy parameter types

olav:


src/component/splash/SplashPageOperators/SplashPageOperators.styles.tsx
```diff
@@ -72,7 +72,7 @@ export const useStyles = makeStyles()(theme => ({
         color: 'inherit',
     },
     button: {
-        background: 'white',
+        background: 'white !important',
```
FredrikOseberg, line 75: Why `!important` ?

FredrikOseberg (approved):
LGTM. Just curious about the `!important`

src/component/splash/SplashPageOperators/SplashPageOperators.styles.tsx
olav: Just to override the MUI styles. There's probably a nice way to do this but I don't think we'll be making more changes to `SplashPageOperators`, so a quick fix made sense.

PR #951, opened 2022-05-05
Refactored feature toggles table

Tymek:
## About the changes
Refactored table toolbar on top to follow more readable React composition.

### Important files
`src/component/feature/FeatureToggleList/FeatureToggleListTable/FeatureToggleListTable.tsx`

## Discussion points
**- is importing all table elements from index in `component/common/Table` OK?**


FredrikOseberg (approved):
LGTM

Tymek:
[rebased after base PR was merged]

src/component/common/Table/TableCell/TableCell.tsx
```diff
@@ -0,0 +1,9 @@
+import { FC } from 'react';
+import { TableCell as MUITableCell, TableCellProps } from '@mui/material';
+import { useStyles } from './TableCell.styles';
+
+export const TableCell: FC<TableCellProps> = ({ ...props }) => {
+    const { classes: styles } = useStyles();
+
+    return <MUITableCell className={styles.tableCell} {...props} />;
```
nunogois, line 8: In cases where we create a class for very few simple styles (like here, where we have a single one) I would argue that it would be easier and clearer to add the styles with the `sx` attribute and deleting the separate `TableCell.styles.ts` file. But I guess it's a matter of preference and this may just be my Tailwind bias. 

Maybe this is something we would want to align on as a team.

src/component/common/Table/TableSearch/TableSearch.styles.ts
```diff
@@ -0,0 +1,59 @@
+import { makeStyles } from 'tss-react/mui';
+
+export const useStyles = makeStyles()(theme => ({
+    fieldWidth: {
```
nunogois, line 4: We might want to rename these classes to something more appropriate. Maybe `searchField`, `searchFieldExpanded` and `searchFieldCollapsed`?

nunogois (approved):

src/component/common/Table/TableCell/TableCell.tsx
FredrikOseberg: Agree, another option is to put the makeStyles file into the same file as the TableCell to avoid cluttering the render function. Though I suspect we will soon need to have a discussion on emotion and the way material-ui is taking their styling.

src/component/common/Table/SortableTableHeader/SortableTableHeader.styles.ts
```diff
@@ -4,6 +4,7 @@ export const useStyles = makeStyles()(theme => ({
     tableHeader: {
         '& > th': {
             border: 0,
+            overflow: 'hidden',
```
olav, line 7: Looks like this hides the table header button focus outline.

src/component/common/Table/TableContainer/TableContainer.tsx
```diff
@@ -0,0 +1,15 @@
+import { forwardRef } from 'react';
+import { Paper, PaperProps } from '@mui/material';
+import { useStyles } from './TableContainer.styles';
+
+export const TableContainer = forwardRef<HTMLDivElement, PaperProps>(
+    ({ children, ...props }, ref) => {
+        const { classes: styles } = useStyles();
```
olav, line 7: The `classes: styles` destructuring was done to make the MUI migration diff smaller. 

Can also use `classes` directly.

src/component/common/Table/TableSearch/TableSearch.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    onChange={onSearchChange}
+                    placeholder={placeholder}
+                    onBlur={onBlur}
+                />
+            </AnimateOnMount>
+            <ConditionallyRender
+                condition={!isSearchExpanded && !isAnimating}
+                show={
+                    <Tooltip title={placeholder} arrow>
+                        <IconButton
```
olav, line 62: Should connect the button to the field through attributes. Could use `aria-expanded`, `aria-controls` etc.

src/component/feature/FeatureToggleList/FeatureToggleListTable/FeatureToggleListTable.tsx
```diff
@@ -139,27 +148,29 @@ export const FeatureToggleListTable: VFC<IExperimentProps> = ({
...
-            }
-        >
+        <TableContainer ref={ref}>
+            <TableToolbar title={`Feature Flags (${data.length})`}>
+                <TableSearch
+                    initialValue={globalFilter}
+                    onChange={setGlobalFilter}
+                />
+                {/*
+                <TableFilter />
```
olav, line 158: Removable?

olav (changes requested):

src/component/common/Table/TableCell/TableCell.tsx
Tymek: `sx` has worse performance. For table cell it's important. https://mui.com/system/basics/#performance-tradeoff

src/component/common/Table/TableSearch/TableSearch.tsx
Tymek: This will require a lot of changes, because component was written with mounting/dismounting elements.

olav (approved):

PR #956, opened 2022-05-06
refactor: add Plausible tracker on SaaS domain

olav:


FredrikOseberg (changes requested):
Super nice, but let's also use a flag for this from uiConfig. That way we can control when / where we turn this on and exclude certain instances if a customer requests it (or if we want to turn off instances like the demo instance or our internal hosted instance).

FredrikOseberg (approved):
LGTM

PR #960, opened 2022-05-06
Theme consistency updates

Tymek:
## About the changes
Minor updates to how table elements look like. Also a refactor of theme, to make it more consistent with MUI.

### Important files
`src/themes/theme.ts`


src/component/styles.module.scss
```diff
@@ -3,11 +3,11 @@
     ul {
         margin: 0;
     }
-    hr {
-        margin: 0;
-        height: auto;
-        border-color: rgba(0, 0, 0, 0.12);
-    }
+    // hr {
```
olav, line 6: Removable?

src/component/user/NewUser/NewUser.tsx
```diff
@@ -73,7 +73,7 @@ export const NewUser = () => {
                                 <div className={styles.roleContainer}>
                                     <ConditionallyRender
                                         condition={Boolean(
-                                            authDetails?.options?.length
+                                            authDetails?.options?.length || true
```
olav, line 76: Removable?

src/hooks/useFeaturesSort.ts
```diff
@@ -175,7 +175,7 @@ const compareNullableDates = (
     a: Date | null | undefined,
     b: Date | null | undefined
 ): number => {
-    return a && b ? a.getTime() - b.getTime() : a ? 1 : b ? -1 : 0;
+    return a && b ? a?.getTime?.() - b?.getTime?.() : a ? 1 : b ? -1 : 0;
```
olav, line 178: Seems like the previous version should work.

olav (changes requested):

src/hooks/useFeaturesSort.ts
Tymek: I had some kind of edge case, where it didn't and broke the UI, similar to the issue with 4.10 bug. It's a temporary fix anyway, since we will migrate this to new table setup this week.

olav (approved):

PR #969, opened 2022-05-09
Disable default env in project view

sighphyre:


src/component/project/ProjectEnvironment/ProjectEnvironment.tsx
```diff
@@ -159,6 +159,7 @@ const ProjectEnvironmentList = ({
                                     env.enabled ? 'Disable' : 'Enable'
                                 } environment`}
                                 size="medium"
+                                disabled={env.name === 'default'}
```
FredrikOseberg, line 162: Hmm. I think this makes sense if it is OSS, but if we have customers on SaaS/self-hosted enterprise then we'd display this one as disabled regardless of the state. I think we'll need to check for that too. There should be an isOss function you can use.

FredrikOseberg (changes requested):
Nice! Just need to account for enterprise here.

src/component/project/ProjectEnvironment/ProjectEnvironment.tsx
sighphyre: Ah yeah, good point. So I've changed this to only disable this if it's the default project and we're on an OSS version. This means that enterprise folks can re enable it if they really want to. Is that closer to what you'd imagined here?

FredrikOseberg (approved):
LGTM from this side. Might want a guard in the API too.

src/component/project/ProjectEnvironment/ProjectEnvironment.tsx
FredrikOseberg: Yeah, some of our enterprise customers may be using the default environment still, if they were using it heavily before we introduced environments. A blanket disable would remove their ability to control the default environment. We might want to deprecate this in v5 but for now we'll have to live with the default environment.

PR #970, opened 2022-05-09
fix: feature toggle list and UI adjustments

nunogois:
## About the changes
https://trello.com/c/iYsYfeNM


src/themes/theme.ts
```diff
@@ -67,6 +67,8 @@ export default createTheme({
         },
         divider: colors.grey[300],
         dividerAlternative: colors.grey[500],
+        hover: 'rgba(120,120,122,0.1)',
+        highlight: '#FFEACC',
```
Tymek, line 71: 👍🏻

src/themes/theme.ts
```diff
@@ -31,6 +31,7 @@ export default createTheme({
     },
     shape: {
         borderRadius: '3px',
+        borderRadiusMedium: '8px',
```
Tymek, line 34: @NicolaeUnleash can we standardize border radius? Does 8px vs 10px make a difference in the UI?

src/themes/themeTypes.ts
```diff
@@ -49,6 +49,22 @@ declare module '@mui/material/styles' {
             abandoned: string;
         };
         dividerAlternative: string;
+        /**
+         * For hover effect
+         */
+        hover: string;
+        /**
+         * For highlight effect
```
Tymek, line 57:
```suggestion
         * Text highlight effect color. Used when filtering/searching over content.
```

src/themes/themeTypes.ts
```diff
@@ -49,6 +49,22 @@ declare module '@mui/material/styles' {
             abandoned: string;
         };
         dividerAlternative: string;
+        /**
+         * For hover effect
```
Tymek, line 53: This can be more descriptive. `hover` of what?

src/themes/theme.ts
NicolaeUnleash: that's a good idea to make a standard, this is what i used to keep consistency, maybe we can create these 3 tokens(s,m,l):
-> 4px (for buttons, inputs)
-> 8px for (table header, dropdown)
-> 12px for bigger elements (bigger containers, dialogs)

not sure what we have now in the app, but it could be a start to align them

src/themes/app.css
```diff
@@ -135,7 +135,11 @@ p {
 }
 
 a {
-    color: inherit;
+    color: #615bc2;
```
Tymek, line 138: Can we avoid this? Is `components.MuiLink.styleOverrides.root` not working?

src/themes/theme.ts
```diff
@@ -46,27 +47,31 @@ export default createTheme({
...
+            light: colors.red[50],
             main: colors.red[700],
             dark: colors.red[800],
         },
         divider: colors.grey[300],
         dividerAlternative: colors.grey[500],
+        tableHeaderHover: colors.grey[400],
+        highlight: '#FFEACC',
+        sidebarContainer: 'rgba(32,32,33, 0.2)',
+        sidebarDivider: 'rgba(255, 255, 255, 0.3)',
```
Tymek, line 74: Is this different than `theme.palette.divider` or `theme.palette.dividerAlternative`?

src/themes/app.css
nunogois: The problem is that not every "link" uses a MUI Link component. It seems like we have a lot of plain anchor tags all over the app, so this has a lot more coverage.

src/component/feature/FeatureToggleList/FeatureToggleListTable/FeatureLinkCell/FeatureLinkCell.styles.ts
```diff
@@ -26,9 +26,5 @@ export const useStyles = makeStyles()(theme => ({
         textDecoration: 'none',
         fontSize: 'inherit',
         display: 'inline-block',
-        maxWidth: '250px',
-        whiteSpace: 'nowrap',
```
FredrikOseberg, line 30: What happens here if the text overflows the container? We can't just cut the text for feature toggles because the name is the id of the feature toggle, and just cutting it off without any tooltip or some way to see the full name won't work. We'll need to use the StringTruncator component or something similar in situations where we have to trim the user input because of length.

src/component/styles.module.scss
```diff
@@ -17,7 +17,7 @@
     margin: 0 auto;
     flex: 1;
     width: 100%;
-    background-color: #ecebeb;
+    background-color: #eaeaed;
```
FredrikOseberg, line 20: Can we replace this module with makeStyles?

FredrikOseberg (changes requested):
Nice. Though we'll need to tackle shortening names differently. This applies generally to overflowing items in the UI, but especially important for feature toggles since that is used as the id for the feature toggle in the SDK. For this PR I think it would be enough to handle it in the feature toggle list and the feature toggle page header.

src/component/feature/FeatureToggleList/FeatureToggleListTable/FeatureLinkCell/FeatureLinkCell.styles.ts
nunogois: That makes sense, however these changes apply to the description, not the name. This is the current behaviour: 
image

This means that the description now wraps to a new line if needed, which also means this is a lot more responsive now, since it behaves properly on all screen sizes. This way we are also able to highlight all parts of the description when searching.

We are aligned on this from a UI/UX perspective, and if the descriptions get too long, we can always truncate them later. Does this make sense for now?

Tymek (approved):
LGTM

PR #978, opened 2022-05-11
fix: resolve issues around changing a toggle's project

olav:


src/utils/sameArrayItems.ts
```diff
@@ -0,0 +1,12 @@
+export const sameArrayItems = <T>(a: T[], b: T[]): boolean => {
```
Tymek, line 1: maybe `arraysHaveSameItems`, or something similar. It's easier to read if you have `is` or `can` or `has` in the name, implying the boolean return type.

src/hooks/api/getters/useProject/useProject.ts
```diff
@@ -1,37 +1,29 @@
...
 import useSort from 'hooks/useSort';
 
 const useProject = (id: string, options: SWRConfiguration = {}) => {
     const { KEY, fetcher } = getProjectFetcher(id);
+    const { data, error, mutate } = useSWR<IProject>(KEY, fetcher, options);
     const [sort] = useSort();
 
-    const { data, error } = useSWR<IProject>(KEY, fetcher, options);
-    const [loading, setLoading] = useState(!error && !data);
+    const refetch = useCallback(() => {
```
Tymek, line 13: I don't think `useCallback` is necessary here, if you assign `refetch: mutate` when returning.

Tymek (approved):

src/hooks/api/getters/useProject/useProject.ts
olav: Right, the purpose of `useCallback` here is to avoid changing the type of the returned refetch.

Maybe a change for later when the hook itself has a proper return type.

FredrikOseberg (approved):
LGTM

PR #985, opened 2022-05-12
feat: add trial expiration warning banner 

olav:


src/utils/formatPath.ts
```diff
@@ -1,51 +1,40 @@
...
-        }
-        basePath = '';
-        return basePath;
-    };
+export const formatApiPath = (path: string, base = basePath): string => {
+    return joinPaths(base, path);
 };
 
-export const getBasePath = getBasePathGenerator();
+export const formatAssetPath = (path: string, base = basePath): string => {
```
FredrikOseberg, line 5: I'm a bit wary of changing the basePath logic. Why was this needed?

FredrikOseberg (changes requested):
I started on this but didn't have time to read through it all. I'll just leave this for the weekend and we can take another look on monday!

src/component/common/InstanceStatus/InstanceStatus.tsx
```diff
@@ -0,0 +1,22 @@
+import { useInstanceStatus } from 'hooks/api/getters/useInstanceStatus/useInstanceStatus';
+import React, { ReactNode } from 'react';
+import { InstanceStatusBar } from 'component/common/InstanceStatus/InstanceStatusBar';
+
+interface IInstanceStatusProps {
+    children: ReactNode;
+}
+
+export const InstanceStatus = ({ children }: IInstanceStatusProps) => {
```
nunogois, line 9: Wouldn't it be clearer to have this as `export const InstanceStatus: FC = ({ children }) => {` and remove the props interface?

src/utils/env.ts
```diff
@@ -0,0 +1,20 @@
+export const isLocalhostDomain = (
+    hostname = window.location.hostname
+): boolean => {
+    return hostname === 'localhost';
```
nunogois, line 4: Maybe some unnecessary returns here, `=> hostname === 'localhost'` feels clear enough to me. Return types can also be inferred without specifying.

src/component/common/InstanceStatus/InstanceStatusBar.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+const calculateTrialDaysRemaining = (
+    instanceStatus: InstanceStatus
+): number | undefined => {
+    return instanceStatus.trialExpiry
+        ? differenceInDays(parseISO(instanceStatus.trialExpiry), new Date())
+        : undefined;
+};
+
+const StyledBar = styled('aside')(({ theme }) => ({
```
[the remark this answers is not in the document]
olav: Yeah, it's the first time we're using their `styled` API. Seems to be where MUI is heading. 

We don't have to start here and now though, could stick to `makeStyles` for now.

src/utils/formatPath.ts
olav: Good point. Needed to run localhost with a basePath to test the instance endpoint, and took the time to refactor a bit.

https://github.com/Unleash/unleash-frontend/blob/trial-status/vite.config.ts#L8

FredrikOseberg (approved):
LGTM

PR #988, opened 2022-05-13
refactor: replace react-dnd with custom implementation

olav:


thomasheartman:
Absolutely unrelated to the changes, but I always read "react-dnd" as "React D&D" (as in Dungeons and Dragons). My life will have a little less magic in it once we remove this package 🧙🏼 🐲  🤪

FredrikOseberg (changes requested):
There seems to be a bit of glitching going on at certain breakpoints around the edges when using the drag-and-drop functionality, could we make it a bit smoother to use? I'd love to lose a dependency, but it feels a bit less smooth in this implementation.

src/component/common/Table/cells/ActionCell/ActionCell.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+interface IContextActionsCellProps {
+    children: ReactNode;
+}
+
+export const ActionCell = ({ children }: IContextActionsCellProps) => {
+    return (
+        <Box
+            data-loading
+            sx={{ display: 'flex', justifyContent: 'flex-end', px: 2 }}
```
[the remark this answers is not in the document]
sighphyre: Nice link! Based on those numbers it still looks pretty fast unless there's rending in a tight loop

src/component/environments/EnvironmentActionCell/EnvironmentActionCell.tsx
```diff
@@ -0,0 +1,191 @@
...
+        useEnvironmentApi();
+
+    const [deleteModal, setDeleteModal] = useState(false);
+    const [toggleModal, setToggleModal] = useState(false);
+    const [confirmName, setConfirmName] = useState('');
+
+    const handleDeleteEnvironment = async () => {
+        try {
+            await deleteEnvironment(environment.name);
+            refetchPermissions();
```
sighphyre, line 54: Just checking here, based on the name this looks like it's a promise. Is this intentionally not awaited?

src/component/environments/EnvironmentActionCell/EnvironmentActionCell.tsx
```diff
@@ -0,0 +1,191 @@
...
+    };
+
+    const handleToggleEnvironmentOn = async () => {
+        try {
+            await toggleEnvironmentOn(environment.name);
+            setToggleModal(false);
+            setToastData({
+                type: 'success',
+                title: 'Project environment enabled',
+                text: 'Your environment is enabled',
```
sighphyre, line 82:
```suggestion
                text: 'Your environment is enabled.',
```

src/component/environments/EnvironmentActionCell/EnvironmentActionCell.tsx
```diff
@@ -0,0 +1,191 @@
...
+            ? handleToggleEnvironmentOff()
+            : handleToggleEnvironmentOn();
+    };
+
+    const handleToggleEnvironmentOn = async () => {
+        try {
+            await toggleEnvironmentOn(environment.name);
+            setToggleModal(false);
+            setToastData({
+                type: 'success',
```
sighphyre, line 80: Not for this PR but at some point is it possible to use an enum/constant/something that a compiler will shout at me for? I spell `success` wrong a lot

FredrikOseberg (approved):
Nice!

src/component/environments/EnvironmentActionCell/EnvironmentActionCell.tsx
```diff
@@ -0,0 +1,191 @@
...
+        ? 'Disable environment'
+        : 'Enable environment';
+
+    const editId = useId();
+    const deleteId = useId();
+
+    return (
+        <ActionCell>
+            <ConditionallyRender
+                condition={updatePermission && !search}
```
sighphyre, line 117: I can't actually tell what the case the boolean logic here describes at a glance. Might be more obvious to folks who have spent more time here but if not, a function that does this with a nice name might be better

src/component/environments/EnvironmentRow/EnvironmentRow.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+interface IEnvironmentRowProps {
+    row: Row;
+    moveListItem: MoveListItem;
+}
+
+export const EnvironmentRow = ({ row, moveListItem }: IEnvironmentRowProps) => {
+    const dragItemRef = useDragItem(row.index, moveListItem);
+    const search = useSearchHighlightContext();
+    const draggable = !search;
```
sighphyre, line 15: searchActive? I want to say this describes if a search is currently being used but it's not quite clear from the variable name?

src/component/environments/EnvironmentTable/EnvironmentTable.tsx
```diff
@@ -0,0 +1,128 @@
...
+                            );
+                        })}
+                    </TableBody>
+                </Table>
+            </SearchHighlightProvider>
+        </PageContent>
+    );
+};
+
+const COLUMNS = [
```
sighphyre, line 104: This is really nice!

sighphyre (approved):
Looks good to me, some small thoughts mostly. Low level logic looks good to me, sounds like @Tymek has looked over as well so I'm happy

src/component/environments/EnvironmentActionCell/EnvironmentActionCell.tsx
olav: These `refetch` functions do not return promises at the moment, but they probably should. Something for later.

sighphyre: Makes sense!

sighphyre: Makes sense!

PR #990, opened 2022-05-16
fix: avoid erasing previous environments data on refetch

olav:


src/component/environments/EnvironmentList/EnvironmentList.tsx
```diff
@@ -52,7 +52,7 @@ const EnvironmentList = () => {
         const item = newEnvList.splice(dragIndex, 1)[0];
 
         newEnvList.splice(hoverIndex, 0, item);
-        refetchEnvironments({ environments: newEnvList }, false);
+        mutateEnvironments({ environments: newEnvList }, false);
```
FredrikOseberg, line 55: Could we encapsulate this in the hook that creates this, so that we don't need to spend time figuring out how mutate interals work when we are in the component level? 

```
const updateEnvironmentsLocally = (environments: IEnvironments[]) => {
   mutate({environments }, false)
}
```

FredrikOseberg (changes requested):

FredrikOseberg (approved):
LGTM

PR #991, opened 2022-05-16
Project features list update

Tymek:


src/component/project/Project/Project.styles.ts
```diff
@@ -36,13 +39,15 @@ export const useStyles = makeStyles()(theme => ({
...
         fontSize: theme.fontSizes.mainHeader,
         fontWeight: 'bold',
         marginBottom: '0.5rem',
-        display: 'grid',
-        gridTemplateColumns: '1fr auto',
+        display: 'flex',
+        justifyContent: 'space-between',
         alignItems: 'center',
-        gridGap: '1rem',
+        gap: '1rem',
```
Tymek, line 50: Is this OK? Browser support is good, but not amazing.

FredrikOseberg: Hmm. How does it look in browser without support? Just ignores the gutters?

src/component/project/Project/Project.tsx
```diff
@@ -116,12 +115,10 @@ const Project = () => {
...
                 <div className={styles.innerContainer}>
-                    <h2
-                        data-loading
-                        className={styles.title}
-                        style={{ margin: 0, width: '100%' }}
-                    >
-                        <div className={styles.titleText}>{project?.name}</div>
+                    <h2 className={styles.title}>
+                        <div className={styles.titleText} data-loading>
+                            {project?.name || projectId}
```
olav, line 120: We should probably not use `projectId` as a fallback for `project.name`, given recent feedback

src/component/project/Project/ProjectOverview.tsx
```diff
@@ -28,7 +28,7 @@ const ProjectOverview = ({ projectId }: IProjectOverviewProps) => {
                     health={health}
                     featureCount={features?.length}
                 />
-                <div className={styles.projectToggles}>
+                <main className={styles.projectToggles}>
```
olav, line 31: There's probably already a `<main>` further up the tree, from `MainLayout`. Should only have one.

src/hooks/usePersistentQuery.ts
```diff
@@ -0,0 +1,20 @@
+import { useNavigate } from 'react-router-dom';
+import { getBasePath } from 'utils/formatPath';
+import useQueryParams from 'hooks/useQueryParams';
+import { usePersistentGlobalState } from './usePersistentGlobalState';
+
+export const usePersistentQuery = (key: string) => {
```
olav, line 6: Does this store the same state in both localStorage and the query string? If so, would one place be enough? 

Also seems unused.

src/hooks/usePersistentGlobalState.ts
```diff
@@ -27,3 +32,32 @@ export const createPersistentGlobalStateHook = <T extends object>(
 
     return () => [container.useGlobalState(key)[0], setGlobalState];
 };
+
+export const usePersistentGlobalState = <T extends object>(
```
olav, line 36: What's the difference between `usePersistentGlobalState` and the existing `createPersistentGlobalStateHook`?

olav (changes requested):

src/component/project/Project/Project.tsx
Tymek: This is so that title takes has a placeholder when loading, that takes space. Otherwise I can add `|| "Project name"`

src/hooks/usePersistentGlobalState.ts
Tymek: I'm experimenting with using SWR for global state instead of `react-hooks-global-state` that was last updated a year ago.

src/component/project/Project/Project.tsx
Tymek: This is so that title takes has a placeholder when loading, that takes space. Otherwise I can add || "Project name"

src/hooks/usePersistentGlobalState.ts
olav: Right, maybe this could be moved to a later PR as well then.

src/component/project/Project/Project.styles.ts
Tymek: Yes. Also, I don't think we should care for browsers from 2020 and back

image

olav (approved):

PR #998, opened 2022-05-18
feat: new contexts table

Tymek:
## About the changes
Redesigned `/context` page. ([Figma](https://www.figma.com/file/qdwpPfuitJUNinm6mvmCmG/Unleash-application?node-id=6970%3A47892))

### Important files
`src/component/common/Table/cells/LinkCell/LinkCell.styles.ts`


src/component/context/ContextList/ContextList.tsx
```diff
@@ -57,103 +148,86 @@ const ContextList: VFC = () => {
...
+                        show={
+                            <TablePlaceholder>
+                                No features or projects found matching &ldquo;
+                                {globalFilter}
+                                &rdquo;
+                            </TablePlaceholder>
+                        }
+                        elseShow={
+                            <TablePlaceholder>
+                                No features available. Get started by adding a
```
FredrikOseberg, line 224: Same here.

src/component/context/ContextList/ContextList.tsx
```diff
@@ -1,42 +1,133 @@
...
     const [showDelDialogue, setShowDelDialogue] = useState(false);
-    const smallScreen = useMediaQuery('(max-width:700px)');
     const [name, setName] = useState<string>();
-    const { context, refetchUnleashContext } = useUnleashContext();
+    const { context, refetchUnleashContext, loading } = useUnleashContext();
     const { removeContext } = useContextsApi();
     const { setToastData, setToastApiError } = useToast();
-    const navigate = useNavigate();
-    const { classes: styles } = useStyles();
+    const data = useMemo(() => {
```
FredrikOseberg, line 37: I think we should talk about these kinds of optimisations. In this case, I think actually using `useMemo` is hurting us more than it's helping us. I don't see how this component will re-render, except perhaps if you trigger useSWRs focus refetch by changing tabs to refetch the context fields from the API. 

But even if this was a component that changed often, how many context fields would users create? 10? 100? Even at those numbers, I think useMemo would still not be worth it. By adding useMemo we are adding complexity, memory usage (memoized data will not be garbage collected), and not gaining a whole lot in the process.  

It _sometimes_ makes sense to use `useMemo`, if the computation is heavy - but there's always a tradeoff and I want us to be conscious about when we use it, and what kind of benefit we get in return. 

This article is a good reference for a discussion: https://kentcdodds.com/blog/usememo-and-usecallback

FredrikOseberg (changes requested):
Cool. Nice to see more and more tables in the new format. A few discussion points to clarify before we merge.

src/component/context/ContextList/ContextList.tsx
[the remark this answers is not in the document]
FredrikOseberg: So in this case I see that this is actually a requirement of react-table: https://react-table.tanstack.com/docs/api/useTable

While it doesn't apply here, I think it's still worth to have a discussion around this to align on future use of this functionality.

FredrikOseberg (approved):
Nice!

PR #999, opened 2022-05-18
Persistent table query

Tymek:
## What
As a user, I expect URL to contain the state of the page. I should be able to share URL to the page with sort order and search query. I also expect to go back to the page that is in the same state as I left it.

## Discussion points
- Are the parameters I chose OK? `sort=<fieldId>` and optional `order=desc`, and search is `q` (search query).
- Should we standardize strict parameter keys in `usePersistentQuery`?
- Working example: [/features?sort=name&order=desc&q=tab](https://unleash-frontend-r6hsgvvsa-unleash-team.vercel.app/features?sort=name&order=desc&q=tab)

## How
image


src/component/feature/FeatureToggleList/FeatureToggleListTable/FeatureToggleListTable.tsx
```diff
@@ -87,20 +90,30 @@ const columns = [
     },
 ];
 
+const defaultSort = { id: 'createdAt', desc: false };
```
olav, line 93: Does this have a type, or could we add one?

olav, line 93: Does this have a type, or could we add one?

src/component/feature/FeatureToggleList/FeatureToggleListTable/FeatureToggleListTable.tsx
```diff
@@ -141,6 +155,24 @@ export const FeatureToggleListTable: VFC<IExperimentProps> = ({
         }
     }, [setHiddenColumns, isSmallScreen, isMediumScreen]);
 
+    useEffect(() => {
+        const tableState: IPageQuery = {};
+        if (sortBy[0]?.id !== defaultSort.id) {
```
FredrikOseberg, line 160: These checks could be clearer. What does `sortBy[0]` refer to? Is it the header field?

src/component/feature/FeatureToggleList/FeatureToggleListTable/FeatureToggleListTable.tsx
```diff
@@ -141,6 +155,24 @@ export const FeatureToggleListTable: VFC<IExperimentProps> = ({
...
+            !(sortBy[0]?.id === defaultSort.id && defaultSort.desc)
+        ) {
+            tableState.order = 'desc';
+        }
+        if (globalFilter) {
+            tableState.q = globalFilter;
+        }
+
+        setQuery(tableState);
+    }, [sortBy, globalFilter]); // eslint-disable-line react-hooks/exhaustive-deps
```
olav, line 174: Do we need the `// eslint-disable-line react-hooks/exhaustive-deps`? 

Would be nice to fix stability issues instead of ignoring these warnings.

olav, line 174: Do we need the `// eslint-disable-line react-hooks/exhaustive-deps`? 

Would be nice to fix stability issues instead of ignoring these warnings.

src/hooks/usePersistentGlobalState.ts
```diff
@@ -31,3 +32,31 @@ export const createPersistentGlobalStateHook = <T extends object>(
...
     return () => [container.useGlobalState(key)[0], setGlobalState];
 };
+
+export const usePersistentGlobalState = <T extends object>(
+    key: string,
+    initialValue: T
+) => {
+    const internalKey = `${key}:usePersistentGlobalState:v1`;
+    const { data, mutate } = useSWR<T>(internalKey, () => {
+        const state = getLocalStorageItem(internalKey) as T;
```
olav, line 42:
```
const state = getLocalStorageItem<T>(internalKey);
```

src/hooks/usePersistentGlobalState.ts
```diff
@@ -31,3 +32,31 @@ export const createPersistentGlobalStateHook = <T extends object>(
 
     return () => [container.useGlobalState(key)[0], setGlobalState];
 };
+
+export const usePersistentGlobalState = <T extends object>(
```
FredrikOseberg, line 36: This one is actually quite clever. I was a bit sceptical at first of using useSWR for this, but if there aren't any drawbacks it will allow us to get rid of a dependency.

src/hooks/usePersistentGlobalState.ts
```diff
@@ -31,3 +32,31 @@ export const createPersistentGlobalStateHook = <T extends object>(
...
+            return state;
+        }
+
+        return initialValue;
+    });
+
+    const output = useMemo(
+        () => (data === undefined ? initialValue : data),
+        [data] // eslint-disable-line react-hooks/exhaustive-deps
+    );
```
olav, line 53: Maybe this could be replaced by `useSWR`s `fallbackData` option.

src/hooks/usePersistentQuery.ts
```diff
@@ -0,0 +1,36 @@
...
+    const navigate = useNavigate();
+    const { data, mutate } = usePersistentGlobalState(
+        `${getBasePath()}:${key}:v1`,
+        Object.fromEntries(params)
+    );
+
+    const setQuery = useCallback<(newData: T, replace?: boolean) => void>(
+        (newData, replace = true) => {
+            if (
+                Object.keys(newData).length === Object.keys(data).length &&
```
FredrikOseberg, line 20: Could be extracted as a helper function for clarity `if (matchesQuery(newData, data)`

src/hooks/usePersistentGlobalState.ts
olav, line 36: Should remove `createPersistentGlobalStateHook`, no point in having both.

src/hooks/usePersistentGlobalState.ts
```diff
@@ -31,3 +32,31 @@ export const createPersistentGlobalStateHook = <T extends object>(
...
+
+    return {
+        data: output,
+        mutate: (value: T) => {
+            setLocalStorageItem(internalKey, value);
+            mutate();
```
olav, line 59: Should this be `mutate(value, false)`? No need to run `fetcher` again.

FredrikOseberg (commented):
I think this makes sense. Regarding discussion points: 
- I think the parameters look sensible
- There might be a benefit having strict parameter keys, we should at least make it difficult to overwrite a key or use the same key

Another discussion point might be if it makes sense to use both localstorage and query parameters, or if we should strive to keep query parameters cross routes (I know some places this will be difficult).

src/component/feature/FeatureToggleList/FeatureToggleListTable/FeatureToggleListTable.tsx
```diff
@@ -22,12 +22,15 @@ import { ConditionallyRender } from 'component/common/ConditionallyRender/Condit
...
 import { PageHeader } from 'component/common/PageHeader/PageHeader';
 import { sortTypes } from 'utils/sortTypes';
+import { usePersistentQuery } from 'hooks/usePersistentQuery';
 
 interface IExperimentProps {
     data: Record<string, any>[];
     isLoading?: boolean;
 }
 
+type IPageQuery = Partial<Record<'q' | 'sort' | 'order', string>>;
```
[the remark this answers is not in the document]
Tymek: I used `q`, because I frequently see it as a parameter. 
- [google.com**/search?q=getunleash](https://www.google.com/search?q=getunleash)
- [stackoverflow.com/search?q=unleash](https://stackoverflow.com/search?q=unleash)
But maybe `?search` is better?
- [en.wikipedia.org/w/index.php?search=feature%20flags](https://en.wikipedia.org/w/index.php?search=feature%20flags)

src/component/feature/FeatureToggleList/FeatureToggleListTable/FeatureToggleListTable.tsx
Tymek: Format from `react-table`. Array, because it can support multi-sort (that we don't use)

src/hooks/api/actions/useApi/useApi.ts
```diff
@@ -40,137 +40,148 @@ const useAPI = ({
...
-            setLoading(true);
-        }
-
-        try {
-            const res = await apiCaller();
-            setLoading(false);
-            if (res.status > 299) {
-                await handleResponses(res, requestId);
-            }
+    const handleResponses = useCallback(
```
Tymek, line 43: git diff didn't pick up on only adding `useCallback` here

src/component/common/Table/SortableTableHeader/CellSortable/SortArrow/SortArrow.styles.ts
```diff
@@ -3,7 +3,8 @@ import { makeStyles } from 'tss-react/mui';
 export const useStyles = makeStyles()(theme => ({
     icon: {
         marginLeft: theme.spacing(0.25),
-        marginRight: -theme.spacing(0.5),
+        // marginRight: -theme.spacing(0.5),
```
FredrikOseberg, line 6: `theme.spacing(-0.5)`?

src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -232,22 +244,61 @@ export const ProjectFeatureToggles = ({
...
+
+            if (searchParams.has('columns')) {
+                const columnsInParams =
+                    searchParams.get('columns')?.split(',') || [];
+                const visibleColumns = [...staticColumns, ...columnsInParams];
+                hiddenColumns = allColumnIds.filter(
+                    columnId => !visibleColumns.includes(columnId)
+                );
+            } else if (storedParams.columns) {
+                console.log('storedParams.columns', storedParams.columns);
```
FredrikOseberg, line 269: Removable.

FredrikOseberg (changes requested):
Nice, though it looks like there are still an issue when trying to select only one column: https://www.loom.com/share/34b4fa1eecb7481e83a4ad27fd80200a

There is also some displacement going on when hovering the headers.

Tymek:
> Nice, though it looks like there are still an issue when trying to select only one column: https://www.loom.com/share/34b4fa1eecb7481e83a4ad27fd80200a
> 
> There is also some displacement going on when hovering the headers.

I fix the column state saving. Displacement is because some columns are centered. Maybe we can figure out something better in the future.

src/component/project/Project/ProjectFeatureToggles/ColumnsMenu/ColumnsMenu.tsx
```diff
@@ -143,6 +151,17 @@ export const ColumnsMenu: VFC<IColumnsMenuProps> = ({
                         />,
                         <MenuItem
                             onClick={() => {
+                                onCustomize([
```
olav, line 154: Could extract `onClick` to a const.

olav (approved):

src/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,65 @@
+import { vi } from 'vitest';
+import { useLocalStorage } from './useLocalStorage';
+import { act, renderHook } from '@testing-library/react-hooks';
+
+describe('useLocalStorage', () => {
```
olav, line 5: Missing `test` in the file name.

src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -268,8 +318,45 @@ export const ProjectFeatureToggles = ({
...
+            tableState.search = filter;
+        }
+        tableState.columns = allColumns
+            .map(({ id }) => id)
+            .filter(
+                id =>
+                    !staticColumns.includes(id) && !hiddenColumns?.includes(id)
+            )
+            .join(',');
+        console.log('tableState.columns', tableState.columns);
```
FredrikOseberg, line 346: Removable.

PR #1015, opened 2022-05-24
Fix: Features table performance

Tymek:
## About the changes
With hundreds of toggles sort and search can be sluggish. This change renders only visible items and some above and below current viewpoint.

## Discussion points
- Should we make `64px` standard height for table rows everywhere?
- Script to seed the db, using `faker.js` https://gist.github.com/Tymek/3db1f7e849c0c62c712317dc6a4a63ad


src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
```diff
@@ -174,6 +190,21 @@ export const FeatureToggleListTable: VFC = () => {
         setStoredParams({ id: sortBy[0].id, desc: sortBy[0].desc || false });
     }, [sortBy, globalFilter, setSearchParams, setStoredParams]);
 
+    const [scrollIndex, setScrollIndex] = useState(0);
+    useEffect(() => {
+        const handleScroll = () => {
+            requestAnimationFrame(() => {
+                const position = window.pageYOffset;
+                setScrollIndex(Math.floor(position / (rowHeight * 5)) * 5);
```
Tymek, line 198: This is a bit obfuscated. "Change value every 5 rows scrolled"

src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
```diff
@@ -91,34 +93,36 @@ export const ProjectFeatureToggles = ({
...
-                                featureEnvironments?.find(
-                                    feature => feature?.name === env
-                                )?.enabled || false,
-                        },
-                    ])
-                ),
-            })
-        );
+        return features
+            .slice(0, features.length > limit ? pageSize : limit)
```
Tymek, line 97: For projects above 300-500 items, there are issues rendering this page. I didn't virtualize it yet, so this is a stop-gap solution.

src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
FredrikOseberg: Could be clearer, yeah.

src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
```diff
@@ -207,14 +238,46 @@ export const FeatureToggleListTable: VFC = () => {
...
-                        {rows.map(row => {
+                    <SortableTableHeader headerGroups={headerGroups} flex />
+                    <TableBody
+                        {...getTableBodyProps()}
+                        style={{
+                            height: `${rowHeight * rows.length}px`,
+                            position: 'relative',
+                        }}
+                    >
+                        {rows.map((row, i) => {
```
FredrikOseberg, line 249: Let's use the full names for parameters, vite will minify for us anyway: `(row, index)`

src/component/project/Project/ProjectFeatureToggles/ProjectFeatureToggles.tsx
FredrikOseberg: This will impact our demo instance. We have over 300 feature toggles there, might not be the best experience for new users, can we paginate instead in the interim?

FredrikOseberg (changes requested):
Some minor details, but this is a lot better than what we have now! 

Note: If you scroll quickly, the screen goes white while it recalculates - I don't think this is a dealbreaker, but I would be very happy if it is possible to avoid it.

src/component/feature/FeatureToggleList/FeatureToggleListTable.tsx
Tymek: Fixed. 

`i` is so universal, that I didn't even thing about it. `e` can stand for Event or Error, and I don't like weird things like `err` or `evt`. I think we have `i` in other places.

PR #1017, opened 2022-05-24
refactor: port health reports to react-table

olav:


src/component/Reporting/ReportTable/ReportTable.tsx
```diff
@@ -0,0 +1,177 @@
...
+        type: report.type,
+        stale: report.stale,
+        status: formatStatus(report),
+        lastSeenAt: report.lastSeenAt,
+        createdAt: report.createdAt,
+        expiredAt: formatExpiredAt(report),
+    };
+};
+
+const COLUMNS = [
```
FredrikOseberg, line 130: I think this should be the default. Cleans up components a lot.

PR #1026, opened 2022-05-25
refactor: port tokens list to react-table

olav:


NicolaeUnleash (changes requested):
@olav
Please can we have these adjustments:

- [x] 1. The sorting to be by default on Created -> newest created on top in case you have a longer list when you do a new it will be on the top of the list (what do you think)
- [x] 2. We need the column "Name" to have the same name as when you create the token -> then it needs to be "Username"
- [ ] 3. The search should highlight the words -> take a look on context fields page
- [x] 4. When no results on search we should have -> There are no results for “keyword” <- (this should be a general solution in all the tables)
- [x] 5. Can we use for "Type" uppercase? -> so it's the same format as in the API command?
- [x] 6. Maybe for the icon to use something else like: "token" or "api" or "key" (the lock is used as action icon in the app and i would not use it here)
- [x] 7. can we also the title like this "API access"

PR #993, opened 2022-05-16
feat: add billing page to admin

nunogois:
## About the changes
https://trello.com/c/fmW7tby2 - Model trial concept - Frontend billing page

### Important files
 - `src/component/admin/billing/Billing.tsx`

## Discussion points
Used a more `styled` and `sx` approach.

PR #1032, opened 2022-05-26
fix: admin routes bug

nunogois:
## About the changes
Fixes a bug that caused flags to be ignored when filtering admin routes, introduced in https://github.com/Unleash/unleash-frontend/pull/993

### Important files
- `Header.tsx`


sighphyre (approved):
Nice catch!

PR #1040, opened 2022-05-27
feat: upgrade users table

nunogois:
## About the changes
https://trello.com/c/BQs64vS2

### Important files
 - `UsersList.tsx`

## Discussion points
With this upgrade, at the end there was a lot of legacy/unused code that looked safely deletable, so I did. 
Let me know if you notice anything wrong, or anything else we should delete that I missed, and I'll gladly include it in this PR.


src/utils/sortTypes.ts
```diff
@@ -7,7 +7,7 @@ export const sortTypes = {
     date: (v1: any, v2: any, id: string) => {
         const a = new Date(v1?.values?.[id] || 0);
         const b = new Date(v2?.values?.[id] || 0);
-        return b?.getTime() - a?.getTime();
+        return a?.getTime() - b?.getTime();
```
nunogois, line 10: This seems to have been the opposite of what we wanted? At least the way I interpret it, that's why I made this change.

src/component/admin/users/UsersList/UsersList.tsx
```diff
@@ -116,136 +97,233 @@ const UsersList = ({ search }: IUsersListProps) => {
...
+                            <Delete />
+                        </PermissionIconButton>
+                    </Box>
+                ),
+                width: 100,
+                disableSortBy: true,
+            },
+        ],
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+        [loading, navigate, isBillingUsers]
```
nunogois, line 211: Was having some trouble with these dependencies. For example, `roles` was suggested as a dependency but when adding it I got an infinite (render?) loop React error. 
These current dependencies make logical sense to me, but maybe someone more experienced with React has a better solution here.

nunogois, line 211: Was having some trouble with these dependencies. For example, `roles` was suggested as a dependency but when adding it I got an infinite (render?) loop React error. 
These current dependencies make logical sense to me, but maybe someone more experienced with React has a better solution here.

olav: Yeah, you were right to add the dependency. Would be good to get rid of all these annotations.

The problem seems to be that the roles object is unstable. `roles` is a new empty array on each render (until the fetch completes). When `roles` change, `columns` change, and then `useTable` probably triggers another render. 

Could change `useUsers` to something like this:

```ts
import useSWR from 'swr';
import { useMemo } from 'react';
import { formatApiPath } from 'utils/formatPath';
import handleErrorResponses from '../httpErrorResponseHandler';

export const useUsers = () => {
    const { data, error, mutate } = useSWR(
        formatApiPath(`api/admin/user-admin`),
        fetcher
    );

    return useMemo(
        () => ({
            users: data?.users || [],
            roles: data?.rootRoles || [],
            loading: !error && !data,
            refetch: () => mutate,
            error,
        }),
        [data, error, mutate]
    );
};

const fetcher = (path: string) => {
    return fetch(path)
        .then(handleErrorResponses('Users'))
        .then(res => res.json());
};
```

src/component/common/Table/cells/TimeAgoCell/TimeAgoCell.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useLocationSettings } from 'hooks/useLocationSettings';
+import { VFC } from 'react';
+import { formatDateYMD } from 'utils/formatDate';
+import { TextCell } from '../TextCell/TextCell';
+import TimeAgo from 'react-timeago';
+
+interface ITimeAgoCellProps {
+    value?: string | number | Date;
+    emptyText?: string;
+    children?: string | number | Date;
```
olav, line 12: Could we remove `children` and just use `value`?

src/component/common/Table/cells/TimeAgoCell/TimeAgoCell.tsx
```diff
@@ -0,0 +1,50 @@
...
+                            date,
+                            locationSettings.locale
+                        )}`}
+                        arrow
+                    >
+                        <Typography noWrap variant="body2" data-loading>
+                            <TimeAgo
+                                date={new Date(date)}
+                                live={false}
+                                title={''}
```
olav, line 42: Can these two be removed?

olav (changes requested):

nunogois:
Now includes these misc UI improvements: https://trello.com/c/oRlfE0St and https://trello.com/c/iINK4ws9

src/component/common/Table/cells/TimeAgoCell/TimeAgoCell.tsx
nunogois: They can, but I don't think they should. 
`live={false}` is our desired behaviour for the most part, but can be a prop. Its default is `true`. 
If we don't set `title={''}` however, we get this UI behaviour where the title shows up along with our tooltip:
image

src/component/admin/users/UsersList/UsersList.tsx
nunogois: Wow, seems to have been exactly that - This works perfectly, thanks! 😄

src/utils/sortTypes.ts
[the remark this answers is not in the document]
Tymek: Can `sortInverted` column property cover your case?
(docs: https://react-table.tanstack.com/docs/api/useSortBy#column-options)

src/component/admin/users/UsersList/UserTypeCell/UserTypeCell.tsx
```diff
@@ -0,0 +1,29 @@
...
+                        <StyledMonetizationOn />
+                    </Tooltip>
+                }
+                elseShow="Free"
+            />
+        </TextCell>
+    );
+};
+
+const StyledMonetizationOn = styled(MonetizationOn)(({ theme }) => ({
```
Tymek, line 26: I prefer to have `styled` higher in the document structure, because it's easier to "scan" what the returned element from the very file is.

src/component/admin/users/UsersList/UsersList.tsx
```diff
@@ -116,136 +97,234 @@ const UsersList = ({ search }: IUsersListProps) => {
...
+                    <TimeAgoCell value={user.seenAt} emptyText="Never logged" />
+                ),
+                sortType: 'date',
+                disableGlobalFilter: true,
+            },
+            {
+                Header: 'Actions',
+                id: 'Actions',
+                align: 'center',
+                Cell: ({ row: { original: user } }: any) => (
```
Tymek, line 172: Maybe a separate component for actions? ~34loc

src/component/admin/users/UsersList/UserTypeCell/UserTypeCell.tsx
nunogois: Thoughts @olav @FredrikOseberg ? Also discussing on Slack, so I'll try to adapt if needed.

nunogois:
Now also includes: https://trello.com/c/tMRWMumW

olav (approved):

PR #1048, opened 2022-05-31
Search keyboard shortcut

Tymek:
`Ctrl+K` / `⌘+K`


src/component/common/Table/TableSearch/TableSearch.tsx
```diff
@@ -16,14 +18,37 @@ interface ITableSearchProps {
...
+    useKeyboardShortcut({ modifiers: ['ctrl'], key: 'f' }, () => {
+        setIsTooltipOpen(true);
+    });
+    useKeyboardShortcut(
+        { modifiers: ['ctrl'], key: 'k', preventDefault: true },
+        () => {
+            setIsSearchExpanded(true);
+        }
+    );
+    useKeyboardShortcut({ key: 'Escape' }, () => {
```
nunogois, line 39: Maybe it could be interesting to also use the same shortcut to close the search? Like a toggle? So close with either `escape` or `ctrl + K` again.

src/hooks/useIsAppleDevice.ts
```diff
@@ -0,0 +1,24 @@
...
+            (
+                navigator as unknown as {
+                    userAgentData: { platform: string };
+                }
+            )?.userAgentData?.platform ||
+            navigator?.platform ||
+            'unknown';
+
+        setIsAppleDevice(
+            platform.includes('Mac') ||
```
nunogois, line 17: I think we need to make a same-case check here, because this is not passing on my Mac, this value is `macOS` for me:
```suggestion
            platform.toLowerCase().includes('mac') ||
```

src/component/common/Table/TableSearch/TableSearch.tsx
```diff
@@ -16,14 +18,37 @@ interface ITableSearchProps {
...
+        if (isSearchExpanded) {
+            setIsSearchExpanded(false);
+        }
+    });
+    const isAppleDevice = useIsAppleDevice();
+    const placeholder =
+        customPlaceholder ??
+        `Search${
+            isAppleDevice !== undefined
+                ? ` (${isAppleDevice ? '⌘' : 'Ctrl'}+K)`
```
nunogois, line 49: I believe this is the wrong character for Ctrl in Mac. It should be something like this instead:
```suggestion
                ? ` (${isAppleDevice ? '⌃' : 'Ctrl'}+K)`
```

nunogois (changes requested):
I also have some concerns regarding UI/UX here, so I'm tagging @NicolaeUnleash

src/component/common/Table/TableSearch/TableSearch.tsx
Tymek: But on mac, "Command+..." shortcuts are standard in system, and there is a check in "useKeyboardShortcut" remapping it

Tymek: I refactored the hook - not it returns the hotkey string :)

src/component/common/Table/TableSearch/TableSearch.tsx
```diff
@@ -16,14 +17,38 @@ interface ITableSearchProps {
...
+                    return true;
+                }
+                if (expanded && !searchInputState) {
+                    return false;
+                }
+                return expanded;
+            });
+        }
+    );
+    useKeyboardShortcut({ modifiers: ['ctrl'], key: 'f' }, () => {
```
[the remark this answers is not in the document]
Tymek: Hotkey is `Ctrl+K`. This part is triggering tooltip to show if you open native browser search :)

FredrikOseberg (approved):
Sweet

src/component/common/Table/TableSearch/TableSearch.tsx
nunogois: Seems like we now have this behaviour:
 - No matter whether we're focused on the field, `escape` toggles the dropdown. If we're focused, it doesn't blur;
 - ⌘+K works well to focus the field, but no longer acts like a toggle - If we're focused, it doesn't blur;

src/component/common/Table/TableSearch/TableSearchField/TableSearchField.tsx
```diff
@@ -20,12 +21,24 @@ export const TableSearchField = ({
...
+        { modifiers: ['ctrl'], key: 'k', preventDefault: true },
+        () => {
+            ref.current?.focus();
+            setShowSuggestions(true);
+        }
+    );
+    useKeyboardShortcut({ key: 'Escape' }, () => {
+        setShowSuggestions(suggestions => !suggestions);
+    });
+    const placeholder = customPlaceholder ?? `Search (${hotkey})`;
```
nunogois, line 41: Does this mean we're only appending the shortcut to the default placeholder, not custom ones from the prop? If so, assuming it would always work, I would argue it makes sense to always append it.

nunogois (approved):
LGTM - But I would still have one last look at UX and the current behaviour after rebasing with @NicolaeUnleash before merging.

PR #1049, opened 2022-05-31
fix: misc ui fixes post tables upgrade

nunogois:
## About the changes
https://trello.com/c/DUedGVC5


olav (approved):

src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
```diff
@@ -1,8 +1,17 @@
+import { styled } from '@mui/material';
 import { Highlighter } from 'component/common/Highlighter/Highlighter';
+import { TextCell } from 'component/common/Table/cells/TextCell/TextCell';
 import { useSearchHighlightContext } from 'component/common/Table/SearchHighlightContext/SearchHighlightContext';
 import { Fragment, VFC } from 'react';
 import { Link } from 'react-router-dom';
 
+const StyledLink = styled(Link)(() => ({
```
Tymek, line 8: If I remember correctly, there is some kind of _hover only_ variant in MUI for this, that works with `:focus` etc.

Tymek (changes requested):

src/component/admin/apiToken/ProjectsList/ProjectsList.tsx
nunogois: It seems that even `LinkCell` is appending specific styles to the `Link` component inside, so I simply added the missing `:focus` behaviour for now.

Tymek (approved):

PR #1050, opened 2022-06-01
fix: always show tokens page header

olav:


src/component/admin/apiToken/ApiTokenTable/ApiTokenTable.tsx
```diff
@@ -69,10 +69,6 @@ export const ApiTokenTable = () => {
         </>
     );
 
-    if (!tokens.length) {
```
ivarconr, line 72: woops!

ivarconr (approved):

Issue #1051, opened 2022-06-01
Project Access tab gives error message if you don't have permission to see users


thomasheartman:
### Describe the bug

When viewing a project I am not a part of and navigating to the project's Access tab, I get an error popup saying "an error occurred while trying to get project access":

![image](https://user-images.githubusercontent.com/17786332/171360156-11402056-1eba-46fc-8c94-2e8739833996.png)

This correlates with a 403 Forbidden response to `/admin/projects/:project/users`.

Even if I don't have access, I can still try and add users. When doing so, the name search works as expected, but the role dropdown has no roles (the dropdown works fine on projects I have access to):

![image](https://user-images.githubusercontent.com/17786332/171360986-9e1ce7a0-30f7-48f0-870e-f4230acc6514.png)

### Steps to reproduce the bug

1. Navigate to a project you don't have access to
2. Select the "access" tab
3. Observe the error popup.

### Expected behavior

If I don't have the right privileges to view users in a project, I'd expect the UI to tell me so in a straightforward manner. More specifically: I would _not_ expect to get an error toast.

Further, if I'm not allowed to add users to a project, I'd expect the input controls to let me know that I can't add a user instead of failing to show me available roles.

### Logs, error output, etc.

```bash
// returned from the users endpoint with a 403 status code
{
	"permission": "UPDATE_PROJECT",
	"message": "You need permission=UPDATE_PROJECT to perform this action"
}
```

### Screenshots

Screenshots are included above.

### Additional context

I've got a feeling this might be related to the recent GDPR changes that were made, where we introduced an opt-in feature to hide users' emails in the UI, but I don't know for sure. 

The changes happen on the demo instance (4.12.3), and I can not reproduce them on a hosted version of Unleash running 4.12.1.

### Unleash version

4.12.3 (the demo instance)

### Hosting type

Hosted by Unleash

PR #1057, opened 2022-06-02
fix: disable project access tab when no access

olav:
https://trello.com/c/OilMJORC/1234-project-access-tab-gives-error-message-if-you-dont-have-permission-to-see-users

Fixes https://github.com/Unleash/unleash-frontend/issues/1051.


FredrikOseberg (approved):
This works. Though you can still visit the URL.

olav:
@FredrikOseberg Went with an alert banner instead to match other pages.

FredrikOseberg (approved):
LGTM, but let's update the text!

PR #1058, opened 2022-06-02
refactor: add OpenAPI schemas to more controllers 

olav:


FredrikOseberg (approved):

FredrikOseberg (changes requested):
Didn't we agree to hold off on implementing openApi fetchers here before we could have the output validation running for a time to assess if there are any databases out of sync before we start implementing it? Or is there a change in how strict this library now parses the JSON response?

olav:
@FredrikOseberg Sure, that makes sense. We can keep these changes in a branch for a while.

PR #1060, opened 2022-06-03
fix: make variant payload text box multiline

olav:


FredrikOseberg (changes requested):
Nice. Could we default the minrows to more than one so that it looks like a textarea and not an input?

FredrikOseberg (approved):

PR #1066, opened 2022-06-06
Archive table

andreas-unleash:
## About the changes

Implemented the archive table
New ActionCell to revive archive toggle
New sort parameter archiveAt

Closes # .


src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,326 @@
...
+    const rowHeight = theme.shape.tableRowHeight;
+    const { classes } = useStyles();
+    const isSmallScreen = useMediaQuery(theme.breakpoints.down('md'));
+    const isMediumScreen = useMediaQuery(theme.breakpoints.down('lg'));
+    const { setToastData, setToastApiError } = useToast();
+
+    const { reviveFeature } = useFeatureArchiveApi();
+
+    const onRevive = (feature: string) => {
+        reviveFeature(feature)
```
FredrikOseberg, line 72: Small thing, but we try to standardize on try/catch for consistency.

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,326 @@
...
+                show={<TablePlaceholder />}
+                elseShow={() => (
+                    <>
+                        <SearchHighlightProvider value={globalFilter}>
+                            <Table {...getTableProps()} rowHeight="standard">
+                                <SortableTableHeader
+                                    headerGroups={headerGroups as any}
+                                />
+                                <TableBody {...getTableBodyProps()}>
+                                    {rows.map((row, index) => {
```
FredrikOseberg, line 197: Could be extracted to a function for readability.

src/component/common/Table/cells/ReviveArchivedFeatureCell/ReviveArchivedFeatureCell.tsx
```diff
@@ -0,0 +1,20 @@
+import { VFC } from 'react';
+import { ActionCell } from '../ActionCell/ActionCell';
+import { Undo } from '@mui/icons-material';
+import { IconButton } from '@mui/material';
+
+interface IReviveArchivedFeatureCell {
+    onRevive: any;
+}
+
+export const ReviveArchivedFeatureCell: VFC<IReviveArchivedFeatureCell> = ({
```
FredrikOseberg, line 10: This is so specific that it should be moved to `/component/archive/ArchiveTable/ReviveArchivedFeatureCell/ReviveArchivedFeatureCell.tsx`

We try to group elements that belong together so that we can cascade deletion of subcomponents/styles if we remove a component.

FredrikOseberg (changes requested):
Nice. Some minor things left.

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,329 @@
...
+        try {
+            await reviveFeature(feature);
+            await refetch();
+            setToastData({
+                type: 'success',
+                title: "And we're back!",
+                text: 'The feature toggle has been revived.',
+                confetti: true,
+            });
+        } catch (e: any) {
```
FredrikOseberg, line 81: Can use `unknown` here and the helper function `formatUnknownError()`

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,329 @@
...
+                            }
+                        />
+                    </>
+                )}
+            />
+        </PageContent>
+    );
+};
+
+function getColumns(onRevive: (feature: string) => Promise<void>) {
```
FredrikOseberg, line 255: Small nit, but we try to standardize where we can and usually use `const getColumns = () =>` as function notation.

src/component/archive/ArchiveTable/ReviveArchivedFeatureCell/ReviveArchivedFeatureCell.tsx
```diff
@@ -0,0 +1,20 @@
+import { VFC } from 'react';
+import { ActionCell } from '../../../common/Table/cells/ActionCell/ActionCell';
+import { Undo } from '@mui/icons-material';
+import { IconButton } from '@mui/material';
+
+interface IReviveArchivedFeatureCell {
+    onRevive: any;
```
FredrikOseberg, line 7: Could also be typed

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,329 @@
...
+        );
+    };
+
+    return (
+        <PageContent
+            isLoading={loading}
+            header={
+                <PageHeader
+                    title={`${
+                        inProject ? 'Project Features Archive' : 'Archived'
```
FredrikOseberg, line 205: Since we have a container for each of these, couldn't the title also be passed in from the outside instead of the `inProject` prop? Then the ArchiveTable truly would keep no internal state knowledge about where it is used.

src/component/archive/ArchiveTable/ReviveArchivedFeatureCell/ReviveArchivedFeatureCell.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const ReviveArchivedFeatureCell: VFC<IReviveArchivedFeatureCell> = ({
+    onRevive,
+}) => {
+    return (
+        <ActionCell>
+            <IconButton onClick={onRevive}>
```
FredrikOseberg, line 15: This needs to be a `<PermissionIconButton />` and take the projectId and permission as props.

FredrikOseberg (changes requested):

src/component/archive/ArchiveTable/ReviveArchivedFeatureCell/ReviveArchivedFeatureCell.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const ReviveArchivedFeatureCell: VFC<IReviveArchivedFeatureCell> = ({
+    onRevive,
+    project,
+}) => {
+    return (
+        <ActionCell>
+            <PermissionIconButton
+                onClick={onRevive}
+                projectId={project}
+                permission={ADMIN}
```
FredrikOseberg, line 21: It's not obvious, but the correct permission here is `UPDATE_FEATURE`

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,329 @@
...
+import { useVirtualizedRange } from '../../../hooks/useVirtualizedRange';
+import {
+    featuresPlaceholder,
+    PageQueryType,
+} from '../../feature/FeatureToggleList/FeatureToggleListTable';
+import theme from 'themes/theme';
+import { FeatureSchema } from '../../../openapi';
+import { useFeatureArchiveApi } from '../../../hooks/api/actions/useFeatureArchiveApi/useReviveFeatureApi';
+import useToast from '../../../hooks/useToast';
+import {formatUnknownError} from "../../../utils/formatUnknownError";
```
FredrikOseberg, line 41: Seems to be some formatting going wrong here

FredrikOseberg (changes requested):
Need to set the correct permission, then this should be good.

src/component/archive/ArchiveTable/ReviveArchivedFeatureCell/ReviveArchivedFeatureCell.tsx
FredrikOseberg: We should also add: `tooltipProps={{ title: 'Revive feature' }}` for accessability here.

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,329 @@
...
+                            }
+                        />
+                    </>
+                )}
+            />
+        </PageContent>
+    );
+};
+
+const getColumns = (onRevive: (feature: string) => Promise<void>) => {
```
Tymek, line 254: This will cause render loop. You need to use `useMemo` and probably wrap `onRevive` in `useCallback`.

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,329 @@
...
+    useTable,
+} from 'react-table';
+import { SearchHighlightProvider } from 'component/common/Table/SearchHighlightContext/SearchHighlightContext';
+import { useMediaQuery } from '@mui/material';
+import { sortTypes } from 'utils/sortTypes';
+import { useEffect, useMemo, useState } from 'react';
+import { HighlightCell } from 'component/common/Table/cells/HighlightCell/HighlightCell';
+import { DateCell } from 'component/common/Table/cells/DateCell/DateCell';
+import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
+import { FeatureTypeCell } from '../../common/Table/cells/FeatureTypeCell/FeatureTypeCell';
```
Tymek, line 25: You can use paths like `'component/` `'hook/` etc

src/hooks/useFeaturesSort.ts
```diff
@@ -15,6 +15,7 @@ export type FeaturesSortType =
     | 'enabled'
     | 'stale'
     | 'created'
+    | 'archived'
```
Tymek, line 18: I think entire `useFeaturesSort` unnecessary. `ArchiveListContainer` is not in use anywhere.

Tymek (changes requested):

src/hooks/useFeaturesSort.ts
FredrikOseberg: To add to this: If there are components not in use as a result of this PR replacing them, it would be good to delete them in the same PR.

src/component/archive/ArchiveTable/ArchiveTable.tsx
andreas-unleash: Tymek was right here - Statically defining works fine if you do not have to pass a function

FredrikOseberg: > Tymek was right here - Statically defining works fine if you do not have to pass a function

Nice catch @Tymek. For my curiosity, would it work if onRevive was a stable reference wrapped in useCallback?

Tymek: Yes, unless `onRevive` had another unstable dependency to it.

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,338 @@
...
+import useToast from 'hooks/useToast';
+import { formatUnknownError } from 'utils/formatUnknownError';
+import { useSearch } from 'hooks/useSearch';
+import { FeatureArchivedCell } from '../../common/Table/cells/FeatureArchivedCell/FeatureArchivedCell';
+import { useVirtualizedRange } from 'hooks/useVirtualizedRange';
+
+export interface IFeaturesArchiveTableProps {
+    archivedFeatures: FeatureSchema[];
+    title: string;
+    refetch: any;
```
Tymek, line 39: This should have static type. Same with `storedParams`, `setStoredParams`, `searchParams` and `setSearchParams`.

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,338 @@
...
+
+    const onRevive = useCallback(async (feature: string) => {
+        try {
+            await reviveFeature(feature);
+            await refetch();
+            setToastData({
+                type: 'success',
+                title: "And we're back!",
+                text: 'The feature toggle has been revived.',
+                confetti: true,
```
Tymek, line 77:
```suggestion
```

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,338 @@
...
+                        onRevive={() => onRevive(original.name)}
+                    />
+                ),
+            },
+            // Always hidden -- for search
+            {
+                accessor: 'description',
+            },
+        ],
+        []
```
Tymek, line 167: ESlint adds warning here

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,338 @@
...
+        setSearchParams(tableState, {
+            replace: true,
+        });
+        setStoredParams({ id: sortBy[0].id, desc: sortBy[0].desc || false });
+    }, [loading, sortBy, searchValue, setSearchParams, setStoredParams]);
+
+    const [firstRenderedIndex, lastRenderedIndex] =
+        useVirtualizedRange(rowHeight);
+
+    const renderRows = () => {
```
Tymek, line 247: If function only returns JSX, it should be in `render`, or separate component.

src/component/common/Table/cells/FeatureArchivedCell/FeatureArchivedCell.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { VFC } from 'react';
+import TimeAgo from 'react-timeago';
+import { Tooltip, Typography } from '@mui/material';
+import { formatDateYMD } from '../../../../../utils/formatDate';
+import { TextCell } from '../TextCell/TextCell';
+import { useLocationSettings } from '../../../../../hooks/useLocationSettings';
```
Tymek, line 6: Definitely there is something wrong with this path.

Tymek (changes requested):

Tymek (approved):

src/component/archive/ArchiveTable/FeatureArchivedCell/FeatureArchivedCell.tsx
```diff
@@ -0,0 +1,36 @@
...
+            {archivedAt && (
+                <Tooltip
+                    title={`Archived on: ${formatDateYMD(
+                        archivedAt,
+                        locationSettings.locale
+                    )}`}
+                    arrow
+                >
+                    <Typography noWrap variant="body2" data-loading>
+                        <TimeAgo date={new Date(archivedAt)} />
```
nunogois, line 30: To make this behave correctly in terms of UI/UX and keep it consistent with other uses of this component, I suggest adding the props: `title="" live={false}`.

src/component/archive/ArchiveTable/FeatureArchivedCell/FeatureArchivedCell.tsx
```diff
@@ -0,0 +1,36 @@
...
+}) => {
+    const { locationSettings } = useLocationSettings();
+
+    if (!archivedAt) return <TextCell />;
+
+    return (
+        <TextCell>
+            {archivedAt && (
+                <Tooltip
+                    title={`Archived on: ${formatDateYMD(
```
nunogois, line 23: I think we removed `on` from tooltips, so maybe this should be `Archived:` instead, but @NicolaeUnleash can help here.

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,360 @@
...
+                        No feature toggles found matching &ldquo;
+                        {searchValue}&rdquo;
+                    </TablePlaceholder>
+                }
+            />
+            <ConditionallyRender
+                condition={rows.length === 0 && searchValue?.length === 0}
+                show={
+                    <TablePlaceholder>
+                        None of the feature toggles where archived yet.
```
nunogois, line 354: Typo on `where` - Should be `were` instead. Suggest aligning with @NicolaeUnleash here just to be sure on the correct text.

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,360 @@
...
+                condition={rows.length === 0 && searchValue?.length > 0}
+                show={
+                    <TablePlaceholder>
+                        No feature toggles found matching &ldquo;
+                        {searchValue}&rdquo;
+                    </TablePlaceholder>
+                }
+            />
+            <ConditionallyRender
+                condition={rows.length === 0 && searchValue?.length === 0}
```
nunogois, line 351: Repetitive `rows.length` check. Could have a wrapper `ConditionallyRender` that would check `rows.length` and then a `ConditionallyRender` inside checking for `searchValue?.length` with a `show` and `elseShow` props. Check the described pattern in `FeatureToggleListTable`.

nunogois (commented):
Even though this is already merged maybe some of these comments can prove useful for a later refactor.

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,360 @@
...
+                              <LinkCell
+                                  title={value}
+                                  to={`/projects/${value}`}
+                              />
+                          ),
+                      },
+                  ]
+                : []),
+            {
+                Header: 'Status',
```
nunogois, line 155: I'm pretty sure this should be `State` instead 🤔

src/component/archive/ArchiveTable/FeatureArchivedCell/FeatureArchivedCell.tsx
NicolaeUnleash: It was like this in my design, i forgot to update. But this will be review and aligned if we implement "X days ago" as a general solution (then we will decide how the tooltips to be)

src/component/archive/ArchiveTable/ArchiveTable.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+    return (
+        <PageContent
+            isLoading={loading}
+            header={
+                <PageHeader
+                    title={`${title} (${
+                        rows.length < data.length
+                            ? `${rows.length} of ${data.length}`
+                            : data.length
```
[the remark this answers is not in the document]
NicolaeUnleash: @nunogois it was more that we need to see how the search and title will work together. if we keep this short version it will work. I will create a task to align this across the entire app

src/component/archive/ArchiveTable/ArchiveTable.tsx
NicolaeUnleash: I have Status in the design, and the design was based on how was implemented, but i need to doublecheck what makes more sense :) Status or State

PR #1069, opened 2022-06-06
fix: constraint-related UI fixes

nunogois:
## About the changes
https://trello.com/c/xHeADCNO

### DIFF
--- a/src/component/common/ConstraintAccordion/ConstraintAccordionView/ConstraintAccordionViewHeader/ConstraintAccordionViewHeader.tsx
+++ b/src/component/common/ConstraintAccordion/ConstraintAccordionView/ConstraintAccordionViewHeader/ConstraintAccordionViewHeader.tsx
@@ -1,11 +1,4 @@
-import StringTruncator from 'component/common/StringTruncator/StringTruncator';
-import {
-    Chip,
-    useMediaQuery,
-    IconButton,
-    Tooltip,
-    styled,
-} from '@mui/material';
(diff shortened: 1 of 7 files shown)


src/component/common/ConstraintAccordion/ConstraintAccordionView/ConstraintAccordionViewHeader/ConstraintAccordionViewHeader.tsx
```diff
@@ -46,46 +80,49 @@ export const ConstraintAccordionViewHeader = ({
             onDelete();
         });
 
+    constraint.contextName = 'verylonglonglonglongcontextname';
```
FredrikOseberg, line 83: This was for testing I presume? Currently is crashing the app on vercel because it's trying to assign to a read only variable when adding a constraint.

FredrikOseberg (changes requested):
This should be ready soon, but we need to fix the read only assignment.

src/component/common/ConstraintAccordion/ConstraintAccordionView/ConstraintAccordionViewHeader/ConstraintAccordionViewHeader.tsx
nunogois: You're right of course, forgot about this.

FredrikOseberg (approved):
LGTM

PR #1080, opened 2022-06-09
fix: allow horizontal scroll in tokens table

olav:


thomasheartman (approved):
Nice one! Looks great to me 🥳

PR #1082, opened 2022-06-09
feat: Align switches in table actions

Tymek:


src/hooks/useDragItem.ts
```diff
@@ -9,24 +9,26 @@ export type MoveListItem = (
...
             ref.current.draggable = true;
             ref.current.style.cursor = 'grab';
             ref.current.dataset.index = String(listItemIndex);
-            return addEventListeners(ref.current, moveListItem);
+            return addEventListeners(ref.current, moveListItem, setIsDragging);
         }
     }, [listItemIndex, moveListItem]);
 
-    return ref;
+    return [ref, isDragging];
```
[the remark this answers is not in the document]
olav: Good question. Maybe we could use a separate hook to track if something's being dragged?

```
export const useDragState = (): boolean => {
    const [state, setState] = useState(false);

    useEffect(() => {
        const onDragStart = () => setState(true);
        const onDrop = () => setState(false);

        document.addEventListener('dragstart', onDragStart);
        document.addEventListener('drop', onDrop);

        return () => {
            document.removeEventListener('dragstart', onDragStart);
            document.removeEventListener('drop', onDrop);
        };
    }, [setState]);

    return state;
};
```

Or maybe a CSS selector would do the trick, if the goal is to style something.

Tymek:
[Align enable/disable/deprecation icons in the list](https://trello.com/c/sYbttIPM/1214-align-enable-disable-deprecation-icons-in-the-list)

src/component/addons/AddonList/ConfiguredAddons/ConfiguredAddons.tsx
```diff
@@ -45,19 +45,24 @@ export const ConfiguredAddons = () => {
...
-                    type: 'success',
-                    title: 'Success',
-                    text: 'Addon state switched successfully',
-                });
-            } catch (error: unknown) {
-                setToastApiError(formatUnknownError(error));
-            }
-        },
+        async (addon: IAddon) =>
+            new Promise<void>(async (resolve, reject) => {
```
olav, line 49: Do we need the `new Promise`?

src/component/environments/EnvironmentTable/EnvironmentActionCell/EnvironmentActionCell.tsx
```diff
@@ -141,7 +126,7 @@ export const EnvironmentActionCell = ({
                     <Tooltip
                         title={
                             environment.protected
-                                ? 'You cannot edit environment'
+                                ? 'You cannot edit protected environment'
```
olav, line 129:
```suggestion
                                ? 'You cannot edit a protected environment'
```

src/component/environments/EnvironmentTable/EnvironmentActionCell/EnvironmentActionCell.tsx
```diff
@@ -169,7 +154,7 @@ export const EnvironmentActionCell = ({
                     <Tooltip
                         title={
                             environment.protected
-                                ? 'You cannot delete environment'
+                                ? 'You cannot delete protected environment'
```
olav, line 157:
```suggestion
                                ? 'You cannot delete a protected environment'
```

src/component/environments/EnvironmentTable/EnvironmentIconCell/EnvironmentIconCell.tsx
```diff
@@ -0,0 +1,47 @@
...
+import AccessContext from 'contexts/AccessContext';
+
+const DragIcon = styled(IconButton)(
+    ({ theme }) => `
+    padding: ${theme.spacing(0, 1, 0, 0)};
+    cursor: inherit;
+    color: ${theme.palette.grey[500]};
+    transition: color 0.2s ease-in-out;
+
+    [data-dragging=true] & {
```
olav, line 17: Could be a prop to `EnvironmentIconCell` instead.

src/component/environments/EnvironmentTable/EnvironmentIconCell/EnvironmentIconCell.tsx
```diff
@@ -0,0 +1,47 @@
...
+        color: ${theme.palette.grey[700]};
```
olav, line 18: Seems like a lot of logic for a slight color change, maybe we should just remove `isDragging` instead.

src/component/strategies/StrategiesList/StrategySwitch/StrategySwitch.tsx
```diff
@@ -0,0 +1,38 @@
...
+    disabled,
+    onToggle,
+}) => {
+    const onClick = () => {
+        onToggle(deprecated);
+    };
+
+    return (
+        <Tooltip
+            title={
```
olav, line 23: Looks like this label gets applied to a span, and not to the input element. 

We should also make the tooltip work for disabled switches somehow.

src/hooks/api/actions/useStrategiesApi/useStrategiesApi.ts
```diff
@@ -1,83 +1,100 @@
...
+                body: JSON.stringify(strategy),
+            });
+
+            try {
+                const res = await makeRequest(req.caller, req.id);
+
+                return res;
+            } catch (e) {
+                throw e;
+            }
```
olav, line 25:
```suggestion
            return makeRequest(req.caller, req.id);
```

olav (changes requested):

src/component/environments/EnvironmentTable/EnvironmentIconCell/EnvironmentIconCell.tsx
Tymek: It's here because `EnvironmentRow` isn't directly connected. It's either this or context.

src/component/strategies/StrategiesList/StrategySwitch/StrategySwitch.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const title = deprecated
+        ? 'Excluded from strategy list'
+        : 'Included in strategy list';
+
+    return (
+        <Tooltip
+            title={disabled ? 'You cannot disable default strategy' : title}
+            arrow
+        >
+            <span>
```
olav, line 30: Looks like `Tooltip` applies a label to the span, but not the input element. Maybe something like this would work:

https://github.com/Unleash/unleash-frontend/blob/main/src/component/feature/FeatureStrategy/FeatureStrategyIcon/FeatureStrategyIcon.tsx#L20-L24

src/component/addons/AddonList/ConfiguredAddons/ConfiguredAddonsActionCell/ConfiguredAddonsActionsCell.tsx
```diff
@@ -23,19 +25,24 @@ export const ConfiguredAddonsActionsCell = ({
...
-                permission={UPDATE_ADDON}
-                onClick={() => toggleAddon(original)}
-                tooltipProps={{ title: 'Toggle addon' }}
-            >
-                <ConditionallyRender
-                    condition={original.enabled}
-                    show={<Visibility />}
-                    elseShow={<VisibilityOff />}
+            <Tooltip title={isEnabled ? 'Disable' : 'Enable'} arrow>
+                <PermissionSwitch
```
olav, line 39: Doesn't look like the `input` gets a label, should fix that somehow. Maybe MUI has a way.

src/component/environments/EnvironmentTable/EnvironmentActionCell/EnvironmentActionCell.tsx
```diff
@@ -108,31 +102,22 @@ export const EnvironmentActionCell = ({
...
-                        <IconButton
-                            onClick={() => setToggleModal(true)}
-                            size="large"
-                        >
-                            <PowerSettingsNew />
-                        </IconButton>
-                    </Tooltip>
+                    <>
+                        <Tooltip title={toggleIconTooltip} arrow>
+                            <PermissionSwitch
```
olav, line 112: Same as above, the input is missing a label. There are some preexisting issues with the button labels on this page as well. I can take a look at those in a separate PR.

olav (changes requested):

olav (approved):

PR #1084, opened 2022-06-10
fix: resolve issues with project edit/delete button conditions

olav:


src/component/project/Project/EditProject/EditProject.tsx
```diff
@@ -68,6 +73,12 @@ const EditProject = () => {
         navigate(-1);
     };
 
+    const accessDeniedAlert = !hasAccess(UPDATE_PROJECT, projectId) && (
+        <Alert severity="error" sx={{ mb: 4 }}>
+            You do not have access to edit this project.
```
thomasheartman, line 78: Maybe this should be permission instead?

```suggestion
            You do not have permission to edit this project.
```

thomasheartman: Or maybe even:

```suggestion
            You do not have the required permissions to edit this project.
```

src/component/project/ProjectList/ProjectList.tsx
```diff
@@ -98,12 +98,12 @@ export const ProjectListNew = () => {
...
-                        memberCount={project?.memberCount ?? 0}
-                        health={project?.health}
-                        id={project?.id}
-                        featureCount={project?.featureCount}
+                        onHover={() => handleHover(project.id)}
+                        name={project.name}
+                        memberCount={project.memberCount ?? 0}
+                        health={project.health}
+                        id={project.id}
+                        featureCount={project.featureCount}
```
thomasheartman, line 106: Oh, this is a nice change. How come we can remove the question marks now? Or where they never necessary?

thomasheartman (commented):
Looks pretty good to me! I have a suggestion and a question inline. I'd also like to know exactly how the 'hidden' buttons work for oss users. Does it mean that they'll be able to see the "edit" screen, but never save anything? Do they get an explanation why they can't save anything?

I also think I may have found a little bug (don't know whether it's new or not): if I try to click the 'delete project' option for the default project, I'm instead taken to the default project page. It's as if the click goes through the dropdown, but even if I click where the dropdown overlaps with the project next to it, I'm still taken to the default project.

Further: are there any explanations why the project can't be deleted? I couldn't find any. It might be confusing to users if they don't understand why its greyed out. And potentially even more so for screen readers or keyboard navigators. 

Also: the disabling apparently doesn't apply to projects that can't be deleted because of active toggles: Is that on purpose? We might not have the required data for that in this view?

olav:
@thomasheartman 

> I'd also like to know exactly how the 'hidden' buttons work for oss users. Does it mean that they'll be able to see the "edit" screen, but never save anything? Do they get an explanation why they can't save anything?

The PR simply hides buttons that would take you to the edit screen. CRUD for projects is only for Enterprise, so I think hiding those buttons for OSS makes sense. The screens could still be accessed directly, but that seems unusual. 

> I also think I may have found a little bug (don't know whether it's new or not): if I try to click the 'delete project' option for the default project, I'm instead taken to the default project page. It's as if the click goes through the dropdown, but even if I click where the dropdown overlaps with the project next to it, I'm still taken to the default project.

Nice catch, @NicolaeUnleash found the same bug. Should be fixed by https://github.com/Unleash/unleash-frontend/pull/1084/commits/cd09a1a8b0933386b248215095353e445e42de37

> Further: are there any explanations why the project can't be deleted? I couldn't find any. It might be confusing to users if they don't understand why its greyed out. And potentially even more so for screen readers or keyboard navigators.

Agreed, would be great with better explanations for why things are disabled, lots of gray buttons. Sounds like a separate task.

> Also: the disabling apparently doesn't apply to projects that can't be deleted because of active toggles: Is that on purpose? We might not have the required data for that in this view?

Yeah we should do something better here. @NicolaeUnleash made a card already: https://trello.com/c/Fcm1c2fO/712-delete-project-with-feature-toggles-message

src/component/project/ProjectCard/ProjectCard.tsx
```diff
@@ -104,9 +119,12 @@ export const ProjectCard = ({
                             e.preventDefault();
                             setShowDelDialog(true);
                         }}
+                        disabled={!canDeleteProject}
                     >
                         <Delete className={classes.icon} />
-                        Delete project
+                        {id === DEFAULT_PROJECT_ID && !canDeleteProject
+                            ? 'Cannot delete default project'
```
thomasheartman, line 126:
```suggestion
                            ? 'You can't delete the default project'
```

I like this change a lot, but I think the wording could be improved a bit. How about this?

Also, will this be obvious to people with assistive tech? You can't focus the disabled button at least, but I'm not sure how that works with screen readers. I guess it announces the list?

thomasheartman (approved):
> The PR simply hides buttons that would take you to the edit screen. CRUD for projects is only for Enterprise, so I think hiding those buttons for OSS makes sense. The screens could still be accessed directly, but that seems unusual.

Got it. When they do get there, though: what happens? Are the buttons there or hidden?

> Nice catch, @NicolaeUnleash found the same bug. Should be fixed by https://github.com/Unleash/unleash-frontend/commit/cd09a1a8b0933386b248215095353e445e42de37

Yup! Thank you 🙌🏼 

> Agreed, would be great with better explanations for why things are disabled, lots of gray buttons. Sounds like a separate task.

Yes! That sounds like a plan.

> Yeah we should do something better here. @NicolaeUnleash made a card already: https://trello.com/c/Fcm1c2fO/712-delete-project-with-feature-toggles-messageYeah we should do something better here. @NicolaeUnleash made a card already: https://trello.com/c/Fcm1c2fO/712-delete-project-with-feature-toggles-message

Siiiick. That's cool then.

-----


I've added another comment/suggestion inline, but when that's taken care of, I'm happy to let this go in based on your responses. I'll approve it for now and assume that you'll handle any outstanding issues 😄

src/component/project/ProjectCard/ProjectCard.tsx
olav: Good question. VoiceOver allows focus and announces the button as "dimmed", which seems appropriate.

thomasheartman: Right, yeah. That sounds about right to me, that. That means non-sighted and sighted users should both be able to see and read the text. So that strikes me as good :tm:

olav:
> Got it. When they do get there, though: what happens? Are the buttons there or hidden?

So, if you go directly to the edit project page you'll get 404 errors, on page load for non-default projects, and on save. 

Maybe it would be better to remove the frontend route altogether for OSS, so that the path is a 404 as well.

thomasheartman:
> So, if you go directly to the edit project page you'll get 404 errors, on page load for non-default projects, and on save.
> 
> Maybe it would be better to remove the frontend route altogether for OSS, so that the path is a 404 as well.

Yeah, that makes sense to me, at least 🤷🏼 If there's nothing they can do there, I think it's more confusing that they can enter the page at all.

thomasheartman:
@olav I just realized that the contrast may not be sufficient for the disabled text. According to firefox: 

image

olav:
@thomasheartman Thanks, should be better now.

PR #1085, opened 2022-06-10
refactor: avoid relative positioning for tbody

olav:


src/component/common/Table/SortableTableHeader/SortableTableHeader.styles.ts
```diff
@@ -1,9 +1,13 @@
 import { makeStyles } from 'tss-react/mui';
 
+export const TABLE_HEADER_HEIGHT = 56;
```
FredrikOseberg, line 3: Should this be themed at a later point?

FredrikOseberg (approved):

PR #1090, opened 2022-06-13
refactor: fix trial expiration calculations

olav:


src/utils/instanceTrial.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    if (!trialExpiry || isPast(trialExpiry)) {
+        return;
+    }
+
+    if (
+        differenceInDays(trialExpiry, new Date()) <=
+        TRIAL_EXPIRATION_WARNING_DAYS_THRESHOLD
+    ) {
+        return formatDistanceToNowStrict(trialExpiry);
```
sighphyre, line 32: I didn't know this function existed, neat! Thanks!

src/utils/instanceTrial.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    ).toEqual(undefined);
+    expect(
+        formatTrialExpirationWarning({
+            plan: InstancePlan.UNKNOWN,
+            state: InstanceState.TRIAL,
+            trialExpiry: addHours(new Date(), 2).toISOString(),
+        })
+    ).toEqual('2 hours');
+    expect(
+        formatTrialExpirationWarning({
```
sighphyre, line 53: So I'm not sure this test data is all that interesting. Some food for thought, assuming my brain can work out dates on the fly and my reading of the docs on formatDistanceToNowStrict is correct (i.e. default is to round over ceil or floor):
- 252 hours prior to expiry (that's 10 and a half days before expiry at 12am) you'll get a message telling you 11 days remain. Should 10 and a half days be represented as 11 days? I don't actually know. But the code does suggest you should only get a message warning of the trial expiry 10 days prior to expiry
- 251 hours prior to expiry (still pretty much 10 and a half days but now 1 pm, you'll get a message telling you that you now have 10 days remaining. I suspect to watch that tick over from 11 days to 10 days over lunch is probably confusing

I don't know if the above is what we expect to see. If it isn't then we should change this. If it is, it would be great to encode those decisions in test data so we can ensure we stick to that decision

I hate dates so much

FredrikOseberg: Since we set the trial start and trial expiry date based on _when_ during the day the customer signs up, I think it makes sense. Then you get 14 whole 24 hour days from when you signed up, which makes the most sense to me.

FredrikOseberg (approved):
Nice

src/utils/instanceTrial.test.ts
olav: Good question, hadn't thought about that. Maybe `floor` makes the most sense here. The banner would show the number of full days you have left, until it starts showing hours. We should probably show too few days instead of too many at least.

sighphyre: @olav I think that probably makes a lot of sense but dates are really hard and users aren't much better than us when it comes to getting confused. Sounds like a problem we should throw at the UX folks. All that being said, this is fixes the thing that would have freaked me out as a user - that I got locked out a day earlier than I expected, displaying 11 vs 10 days early on is a detail. I think this is great and we should go with it and maybe look at this later. I realise my original comment suggested this needed to be fixed but really it was more food for thought that was badly phrased

@FredrikOseberg agreed, small details, this is great, huge improvement over what we had

sighphyre (approved):

PR #1091, opened 2022-06-13
refactor: improve button label markup

olav:


thomasheartman:
Ooh, this looks real nice to me! 🤩 Out of curiosity: could you give a brief summary of the changes and why it improves the markup? For instance, the difference between "aria-describedby" and "aria-labelledby" and such?

olav:
@thomasheartman Sure, updated the description!

src/component/strategies/StrategiesList/StrategiesList.tsx
```diff
@@ -265,81 +270,105 @@ export const StrategiesList = () => {
...
+                    >
+                        <Delete />
+                    </PermissionIconButton>
+                }
+                elseShow={
+                    <Tooltip
+                        title="You cannot delete a built-in strategy"
+                        arrow
+                    >
+                        <div id={id} role="tooltip">
```
thomasheartman, line 358: Hmm, I might be understanding [MDN's guide to the tooltip role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tooltip_Role#associated_wai-aria_roles_states_and_properties), but it says:

> The element that serves as the tooltip container has role="tooltip" set.

Shouldn't the `role=tooltip` go on the `Tooltip` component, then? (full disclosure: I haven't checked the Tooltip component code).

MDN's example shows something like this, at least:

```html
<label for="password">Password:</label>
<input aria-describedby="passwordrules" id="password" type="password" />
<div role="tooltip" id="passwordrules">
  <p>
      Password Rules:
  </p>
  <ul>
    <li> Minimum of 8 characters</li>
    <li> Include at least one lowercase letter, one uppercase letter, one number and one special character</li>
    <li>Unique to this website</li>
  </ul>
</div>
```

src/component/common/PermissionButton/PermissionButton.tsx
```diff
@@ -68,11 +68,11 @@ const PermissionButton: React.FC<IPermissionButtonProps> = ({
             title={formatAccessText(access, tooltipProps?.title)}
             arrow
         >
-            <span id={id}>
+            <span id={id} role="tooltip">
                 <Button
                     onClick={onClick}
                     disabled={disabled || !access}
-                    aria-describedby={id}
+                    aria-labelledby={id}
```
thomasheartman, line 75: Can you reference a tooltip with "aria-labelledby"? According  to [the states and properties section of MDN's docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tooltip_Role#associated_wai-aria_roles_states_and_properties):

> The element that triggers the tooltip references the tooltip element with [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby).

src/component/strategies/StrategiesList/StrategiesList.tsx
olav: Ah you're right, the role is for the container. :+1:

src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
```diff
@@ -40,7 +40,7 @@ export const FeatureStrategyMenu = ({
                 projectId={projectId}
                 environmentId={environmentId}
                 onClick={onClick}
-                aria-describedby={popoverId}
+                aria-labelledby={popoverId}
```
thomasheartman, line 43: Are there cases where the `popoverId` being undefined can cause problems? Does the popover open as soon as (or before) the item gets focus? I don't know at all, but is it possible that if the popup doesn't open until after the button gets focus, then assistive tech fails to find the label because it's `undefined`?

src/component/feature/FeatureStrategy/FeatureStrategyIcon/FeatureStrategyIcon.tsx
```diff
@@ -18,7 +18,7 @@ export const FeatureStrategyIcon = ({
     return (
         <StyledIcon>
             <Tooltip title={formatStrategyName(strategyName)} arrow>
-                <div id={id} role="tooltip">
+                <div id={id}>
```
thomasheartman, line 21: What's happening here? Are you saying that the Icon is labelled by a div that's otherwise empty? Or is there some code in the `Icon` component that describes the icon's purpose?

src/component/strategies/StrategiesList/StrategiesList.tsx
```diff
@@ -265,81 +269,105 @@ export const StrategiesList = () => {
...
+                        title="You cannot delete a built-in strategy"
+                        arrow
+                    >
+                        <div id={id}>
+                            <IconButton
+                                disabled
+                                size="large"
+                                aria-labelledby={id}
+                            >
+                                <Delete titleAccess="Delete strategy" />
```
thomasheartman, line 363: In cases where it's disabled, do we announce that to assistive tech? Should the `titleAccess` instead be "you cannot delete a built-in strategy" or is that announced in some other way?

thomasheartman (commented):
Looks a lot better now; thanks for looking into that 🙌🏼  I've still got a couple questions, though: added them to the code.

src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
olav: Looks like my Find/Replace went too far. This should stay `aria-describedby` according to the MUI docs.

https://mui.com/material-ui/react-popover/#basic-popover

Not sure how well it manages focus, could be worth looking into.

src/component/strategies/StrategiesList/StrategiesList.tsx
thomasheartman: This change seems to have disappeared (I'm guessing you removed it), but I'm still curious about how this acts. Do you know?

src/component/feature/FeatureStrategy/FeatureStrategyMenu/FeatureStrategyMenu.tsx
thomasheartman: Got it. Is that something you wanna do or would you rather table it for now? ☺️

olav: Should proabably be a separete PR at least. I'll make a card.

olav: https://trello.com/c/dOUQG85N/722-ensure-that-popovers-focus-handling-works-as-expected

src/component/strategies/StrategiesList/StrategiesList.tsx
olav: Yeah, there were extensive changes to these components in main, also partly focused on a11y.

olav: Combined the labels into one for the disabled buttons. Should be better, though probably still improvable.

thomasheartman (approved):
Yeah, I'm happy! Thanks for your hard work and your clarifications 😄

PR #1107, opened 2022-06-20
refactor: fix handling of expired/churned trial states

olav:


chriswk (approved):
Looks really good, even got a laugh from the self-explaining variable names for the components :)

PR #1117, opened 2022-06-25
fix: copy feature double validation popup

nunogois:
https://trello.com/c/G187Oycd

### DIFF
--- a/src/component/feature/CopyFeature/CopyFeature.tsx
+++ b/src/component/feature/CopyFeature/CopyFeature.tsx
@@ -114,7 +114,6 @@ export const CopyFeatureToggle = () => {
                         variant="outlined"
                         size="small"
                         inputRef={inputRef}
-                        required
                     />
                     <FormControlLabel
                         control={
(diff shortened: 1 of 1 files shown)


src/component/feature/CopyFeature/CopyFeature.tsx
```diff
@@ -114,7 +114,6 @@ export const CopyFeatureToggle = () => {
                         variant="outlined"
                         size="small"
                         inputRef={inputRef}
-                        required
```
olav, line 117: Could perhaps replace `required` with `aria-required` to retain some semantics.

Looks like the removal allows the form to be submitted if the field is never focused.

olav (changes requested):

src/component/feature/CopyFeature/CopyFeature.tsx
nunogois: Would something like https://github.com/Unleash/unleash-frontend/pull/1117/commits/8edd7066cc09df47c6d4768ada9667fce2b16afb make more sense?

olav (approved):

PR #1119, opened 2022-06-27
fix: wrap long words in toast text

olav:
https://trello.com/c/nSzY9ekY/1338-long-string-name-needs-to-wrap-in-toast-messages


sighphyre (approved):
LG

PR #1129, opened 2022-07-05
Playground form

Tymek:
## About the changes
Adds top part of UI for playground feature - for selecting environment and projects, and setting context.

### Important files
`src/component/playground/Playground/Playground.tsx`

### DIFF
--- a/src/component/playground/Playground/Playground.tsx
+++ b/src/component/playground/Playground/Playground.tsx
@@ -1,4 +1,4 @@
-import { FormEventHandler, useMemo, useState, VFC } from 'react';
+import { FormEventHandler, useState, VFC } from 'react';
 import {
     Box,
     Button,
--- a/src/component/playground/Playground/Playground.tsx
+++ b/src/component/playground/Playground/Playground.tsx
@@ -4,13 +4,13 @@ import {
     Button,
     Divider,
     Paper,
-    TextField,
     Typography,
     useTheme,
 } from '@mui/material';
 import { PageContent } from 'component/common/PageContent/PageContent';
 import { PageHeader } from 'component/common/PageHeader/PageHeader';
 import { PlaygroundConnectionFieldset } from './PlaygroundConnectionFieldset/PlaygroundConnectionFieldset';
+import { PlaygroundCodeFieldset } from './PlaygroundCodeFieldset/PlaygroundCodeFieldset';
 
 interface IPlaygroundProps {}
 
@@ -27,7 +27,7 @@ export const Playground: VFC<IPlaygroundProps> = () => {
                     background: theme.palette.grey[200],
                 }}
             >
-                <Box component="form">
+                <Box component="form" sx={{ position: 'relative' }}>
                     <Typography
                         sx={{
                             mb: 3,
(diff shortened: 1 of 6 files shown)


src/component/playground/Playground/Playground.tsx
```diff
@@ -0,0 +1,83 @@
...
+                <Box component="form" onSubmit={onSubmit}>
+                    <Typography
+                        sx={{
+                            mb: 3,
+                        }}
+                    >
+                        Configure playground
+                    </Typography>
+                    <PlaygroundConnectionFieldset />
+                    <Divider
```
FredrikOseberg, line 50: Could this be a `StyledDashedDivider`, then we could reuse it later and we wouldn't have to repeat the styles beneath. Might clean up the render function a bit.

src/component/playground/Playground/PlaygroundCodeFieldset/PlaygroundCodeFieldset.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    {
+                        ...currentValue,
+                        [contextField]: contextValue,
+                    },
+                    null,
+                    2
+                )
+            );
+        } catch (error) {
+            // Shouldn't happened, because button is disabled when there is an error
```
FredrikOseberg, line 81: So if I understand this correctly, there will be a delay where the button will be functional before the error is updated. Should we give any feedback if that is the case or is the button being disabled enough feedback?

FredrikOseberg (changes requested):

src/component/playground/Playground/Playground.tsx
Tymek: I'll add a variant to theme.

src/component/playground/Playground/Playground.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const [context, setContext] = useState<string>();
+    const [contextObject, setContextObject] = useState<string>();
+    const onSubmit: FormEventHandler<HTMLFormElement> = event => {
+        event.preventDefault();
+
+        try {
+            setContextObject(
+                JSON.stringify(JSON.parse(context || '{}'), null, 2)
+            );
+        } catch (e) {}
```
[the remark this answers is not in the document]
Tymek: toast message added

src/component/playground/Playground/Playground.tsx
Tymek: First divider is dashed, the second one isn't

src/component/playground/Playground/PlaygroundCodeFieldset/PlaygroundCodeFieldset.tsx
```diff
@@ -0,0 +1,166 @@
...
+}
+
+export const PlaygroundCodeFieldset: VFC<IPlaygroundCodeFieldsetProps> = ({
+    value,
+    setValue,
+}) => {
+    const theme = useTheme();
+    const { setToastData } = useToast();
+    const { context } = useUnleashContext();
+    const contextOptions = useMemo(
```
FredrikOseberg, line 37: Is this resolved?

FredrikOseberg (changes requested):
Just need a clarification here, but this should be good to go afterwards.

FredrikOseberg (approved):
LGTM!

PR #1136, opened 2022-07-13
Refine Playground

Tymek:
## About the changes
Integrate with API, fix results table, add URL search query parameters.

### DIFF
--- a/src/component/playground/Playground/Playground.tsx
+++ b/src/component/playground/Playground/Playground.tsx
@@ -27,13 +27,12 @@ export const Playground: VFC<IPlaygroundProps> = () => {
     const [environment, onSetEnvironment] = useState<string>('');
     const [projects, onSetProjects] = useState<string[]>([]);
     const [context, setContext] = useState<string>();
-    const [loading, setLoading] = useState(false);
     const [results, setResults] = useState<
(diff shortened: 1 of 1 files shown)


src/component/playground/Playground/Playground.tsx
```diff
@@ -13,37 +14,112 @@ import { PlaygroundConnectionFieldset } from './PlaygroundConnectionFieldset/Pla
...
 
 interface IPlaygroundProps {}
 
 export const Playground: VFC<IPlaygroundProps> = () => {
     const theme = useTheme();
+    const [environment, onSetEnvironment] = useState<string>('');
+    const [projects, onSetProjects] = useState<string[]>([]);
     const [context, setContext] = useState<string>();
-    const [contextObject, setContextObject] = useState<string>();
+    const [loading, setLoading] = useState(false);
```
FredrikOseberg, line 30: Could we use the `loading` property that comes from the `usePlaygroundApi`. We get that state for free and it's connected to the API call.

FredrikOseberg (approved):
Looks good. Small comment to consider.

andreas-unleash (approved):
LGTM!

PR #1139, opened 2022-07-18
feat/playground-second-iteration

FredrikOseberg:
This contains fixes and improvements to the new playground feature:
* The playground will now search for feature toggles automatically if the URL contains a valid query
* The playground defaults to the first available valid environment
* Moved context form above the JSON field
* Fixed a bug where placeholder would display in search
* Changed the layout to keep the results on the side
* Added date input for currentTime and legal values dropdown
* Added sticky form
* Added variant popover information
* Added ability to search for isEnabled result
* Added CodeMirror JSON editor

### DIFF
--- a/src/component/common/Search/Search.tsx
+++ b/src/component/common/Search/Search.tsx
@@ -17,6 +17,7 @@ interface ISearchProps {
     hasFilters?: boolean;
     disabled?: boolean;
     getSearchContext?: () => IGetSearchContextOutput;
+    styles?: React.CSSProperties;
 }
 
 export const Search = ({
@@ -59,7 +60,7 @@ export const Search = ({
     const placeholder = `${customPlaceholder ?? 'Search'} (${hotkey})`;
 
     return (
-        <div className={styles.container}>
+        <div className={styles.container} styles={styles}>
             <div
                 className={classnames(
                     styles.search,
--- a/src/component/common/Search/Search.tsx
+++ b/src/component/common/Search/Search.tsx
@@ -17,7 +17,7 @@ interface ISearchProps {
     hasFilters?: boolean;
     disabled?: boolean;
     getSearchContext?: () => IGetSearchContextOutput;
-    styles?: React.CSSProperties;
+    containerStyles?: React.CSSProperties;
 }
 
 export const Search = ({
@@ -28,6 +28,7 @@ export const Search = ({
     hasFilters,
     disabled,
     getSearchContext,
+    containerStyles,
 }: ISearchProps) => {
     const ref = useRef<HTMLInputElement>();
     const { classes: styles } = useStyles();
@@ -60,7 +61,7 @@ export const Search = ({
     const placeholder = `${customPlaceholder ?? 'Search'} (${hotkey})`;
 
     return (
-        <div className={styles.container} styles={styles}>
+        <div className={styles.container} style={containerStyles}>
             <div
                 className={classnames(
                     styles.search,
--- a/src/component/playground/Playground/Playground.tsx
+++ b/src/component/playground/Playground/Playground.tsx
@@ -1,6 +1,6 @@
 import { FormEventHandler, useEffect, useState, VFC } from 'react';
 import { useSearchParams } from 'react-router-dom';
-import { Box, Paper, useMediaQuery, useTheme, IconButton } from '@mui/material';
+import { Box, Paper, useMediaQuery, useTheme } from '@mui/material';
 import { PageContent } from 'component/common/PageContent/PageContent';
 import { PageHeader } from 'component/common/PageHeader/PageHeader';
 import useToast from 'hooks/useToast';
@@ -11,9 +11,12 @@ import { usePlaygroundApi } from 'hooks/api/actions/usePlayground/usePlayground'
 import { PlaygroundResponseSchema } from 'hooks/api/actions/usePlayground/playground.model';
 import { useEnvironments } from 'hooks/api/getters/useEnvironments/useEnvironments';
 import { PlaygroundForm } from './PlaygroundForm/PlaygroundForm';
-import { resolveDefaultEnvironment, resolveProjects } from './playground.utils';
+import {
+    resolveDefaultEnvironment,
+    resolveProjects,
+    resolveResultsWidth,
+} from './playground.utils';
 import { PlaygroundGuidance } from './PlaygroundGuidance/PlaygroundGuidance';
-import { Help } from '@mui/icons-material';
 import { PlaygroundGuidancePopper } from './PlaygroundGuidancePopper/PlaygroundGuidancePopper';
 
 export const Playground: VFC<{}> = () => {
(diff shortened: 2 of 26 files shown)


src/component/playground/Playground/Playground.tsx
```diff
@@ -103,88 +106,115 @@ export const Playground: VFC<IPlaygroundProps> = () => {
...
+        }
+
+        if (results && !matches) {
+            return '65%';
+        }
+
+        return '50%';
+    };
+
+    const resolveLayout = () => {
```
Tymek, line 155: I think this can be inlined. `sx={{ display: 'flex', flexDirection: !matches ? 'row' : 'column' }}`

src/component/playground/Playground/PlaygroundForm/PlaygroundCodeFieldset/PlaygroundEditor/PlaygroundEditor.tsx
```diff
@@ -0,0 +1,133 @@
...
+interface IPlaygroundEditorProps {
+    context: string | undefined;
+    setContext: Dispatch<SetStateAction<string | undefined>>;
+    error: string | undefined;
+}
+
+const StyledEditorHeader = styled('aside')(({ theme }) => ({
+    height: '50px',
+    backgroundColor: theme.palette.grey[100],
+    borderTopRightRadius: '5px',
```
Tymek, line 18: border radius has values in `theme`
(`theme.shape.`, borderRadius, borderRadiusMedium, borderRadiusLarge, borderRadiusExtraLarge)

FredrikOseberg: Thanks!

src/component/playground/Playground/Playground.tsx
FredrikOseberg: Agree

src/component/common/Search/Search.tsx
```diff
@@ -59,7 +61,7 @@ export const Search = ({
     const placeholder = `${customPlaceholder ?? 'Search'} (${hotkey})`;
 
     return (
-        <div className={styles.container}>
+        <div className={styles.container} style={containerStyles}>
```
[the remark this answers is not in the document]
FredrikOseberg: I do agree, and I'm not entirely happy with how it is now in terms of Styled components / sx / style mix in the project now. But I find using classes is difficult now because (a) we'd like to move away from makeStyles since that is now a third party package that may or may not be maintained in the future, and (b) we need our styles to be connected to the theme, which we can't access through regular styles sheets or css modules.

src/component/playground/Playground/Playground.tsx
```diff
@@ -103,88 +106,114 @@ export const Playground: VFC<IPlaygroundProps> = () => {
...
+        }
+
+        if (results && !matches) {
+            return '65%';
+        }
+
+        return '50%';
+    };
+
+    const formWidth = resolveFormWidth();
```
Tymek, line 155:
```suggestion
    const formWidth = results && !matches ? '35%' : 'auto';
```

Tymek: Same with `resultsWidth`. I think this can be less verbose.

Tymek (commented):
I don't think helper functions are needed for `formWidth`. Also, `borderRadius` from theme wasn't updated.

src/component/playground/Playground/Playground.tsx
FredrikOseberg: I agree. Refactored formWidth to use ternary. I tried a double ternary for resultsWidth but I found it really hard to read so instead I parameterized the helper and moved it out of the way of the component.

Tymek (approved):
LGTM

src/component/playground/Playground/PlaygroundGuidance/PlaygroundGuidance.tsx
```diff
@@ -0,0 +1,44 @@
+import { Typography, Box, Divider } from '@mui/material';
+import { PlaygroundGuidanceSection } from './PlaygroundGuidanceSection/PlaygroundGuidanceSection';
+
+export const PlaygroundGuidance = () => {
+    return (
+        <Box sx={{ ml: 4 }}>
+            <Typography variant="body1">
+                Unleash playground is for helping you to undestand how unleash
+                works, how feature toggles are evaluated and for you to easily
+                debug your feature toggles.
```
thomasheartman, line 10: How about something like this? You might need to adjust the indentation

```suggestion
		The Unleash playground was designed to help you understand how Unleash
		works. You can use it to debug your current feature toggle setup or to
		experiment with how application context impacts feature toggle
		evaluation.
```

src/component/playground/Playground/PlaygroundGuidance/PlaygroundGuidance.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+            <Divider sx={{ mt: 2, mb: 2 }} />
+
+            <Typography variant="body1" sx={{ mb: 1 }}>
+                What you need to do is:
+            </Typography>
+
+            <PlaygroundGuidanceSection
+                headerText="  Select in which environment you want to test your
+                            feature toggle configuration"
```
thomasheartman, line 21:
```suggestion
                headerText="Select an environment to test your
                            feature toggle configuration in"
```

thomasheartman (changes requested):
Nice work! I suggest a few minor text changes

src/component/playground/Playground/PlaygroundGuidance/PlaygroundGuidance.tsx
```diff
@@ -0,0 +1,44 @@
...
+                bodyText="  You can also specify specific projects, or check
+                            toggles in all projects."
+                sectionNumber="1"
+            />
+
+            <PlaygroundGuidanceSection
+                headerText="Select a context field that you'd like to check"
+                bodyText="Configure multiple context fields or leave empty to test against an empty context"
```
thomasheartman, line 29: When you test against an empty context, we automatically apply `appName`, though. Should that be mentioned?

```suggestion
                bodyText="You can configure as many context fields context fields as you want. You can also leave the context empty to test against an empty context."
```

src/component/playground/Playground/PlaygroundGuidance/PlaygroundGuidance.tsx
```diff
@@ -0,0 +1,44 @@
...
+                sectionNumber="2"
+            />
+
+            <PlaygroundGuidanceSection
+                headerText="Try configuration"
```
thomasheartman, line 34:
```suggestion
                headerText="Submit the form to try the configuration"
                bodyText="The results of evaluating your feature toggles will appear after you submit the form."
```

src/component/playground/Playground/PlaygroundGuidance/PlaygroundGuidance.tsx
```diff
@@ -0,0 +1,44 @@
...
+                sectionNumber="3"
+            />
+
+            <PlaygroundGuidanceSection
+                headerText="View the results"
+                sectionNumber="4"
+            />
```
thomasheartman, line 41: Suggest we move this into the previous step. It's not really a step per se.

thomasheartman (changes requested):
A few more text suggestions for the guidance. Think this might be clearer 💁🏼

PR #1145, opened 2022-07-22
feat: lazy load playground

FredrikOseberg:
* Adds lazy loading for playground


thomasheartman (approved):
Ooh, is suspense finally here? 👀  Looks good to me! Changes look pretty basic , so I'm guessing everything just happens magically outside via `lazy`?

FredrikOseberg:
> Ooh, is suspense finally here? 👀 Looks good to me! Changes look pretty basic , so I'm guessing everything just happens magically outside via `lazy`?

Yup, the build tool takes care of splitting it into chunks and lazy dynamically loads it and shows the suspense fallback while loading!

PR #1146, opened 2022-07-22
Feat/grouping fixes

sjaanus:
1. Fixed dataflow error, where group object was not transformed properly
2. Removed deprecated user access endpoints and start using new access endpoint


sjaanus:
Postponed a little, because we probably will need to use 2 separate endpoints instead.

PR #1148, opened 2022-07-25
fix: truncator

FredrikOseberg:
Checks if we have a text before checking the length.

### DIFF
--- a/src/component/common/StringTruncator/StringTruncator.tsx
+++ b/src/component/common/StringTruncator/StringTruncator.tsx
@@ -17,7 +17,7 @@ const StringTruncator = ({
 }: IStringTruncatorProps) => {
     return (
         <ConditionallyRender
-            condition={text?.length > maxLength}
+            condition={(text?.length ?? 0) > maxLength}
             show={
                 <Tooltip title={text} arrow>
                     <span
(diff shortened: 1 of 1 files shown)


src/component/common/StringTruncator/StringTruncator.tsx
```diff
@@ -17,7 +17,7 @@ const StringTruncator = ({
 }: IStringTruncatorProps) => {
     return (
         <ConditionallyRender
-            condition={text.length > maxLength}
+            condition={text?.length > maxLength}
```
thomasheartman, line 20: Nice and good catch! My one suggestion is that we may want to use a default value (0) in case we don't have a value. It seems that comparing `undefined` to number will always return `false`, and that comparing `null` to a number _almost_ treats it as if it's 0 (`null < -1` is `true`, `null > 1` is true, but `null == 0` is `false`).

I think we can count on that being a stable API, but by using a default value instead, we both guard against future changes to this _and_ arguably make our intent clearer. And, on a personal level, comparing two incomparable data types, feels a bit iffy 💁🏼

```suggestion
            condition={(text?.length ?? 0) > maxLength}
```

thomasheartman (approved):
Nice! 🥳 Looks like a good change to me! Approved 🚂  I've only got one minor suggestion here, which you can decide to accept or not.

PR #1149, opened 2022-07-25
fix: pro flow stripe integration refinements

nunogois:
Addresses some changes needed after internal evaluation and testing:
 - Addresses should always be requested;
 - Route title;
 - Text changes;
 - Leverage new `instance/refresh` endpoint for a better UX when signing up;


chriswk (approved):
LGTM :+1:

PR #1151, opened 2022-07-26
refactor: create user avatar component, clean up

nunogois:
## About the changes
https://linear.app/unleash/issue/2-37/avatar-refactor-into-common-component
 > Currently we have a lot of slightly different implementations/stylings of Avatar. Would be nice to refactor it into a common component to be used everywhere.

### Important files

 - `src/component/common/UserAvatar/UserAvatar.tsx`


chriswk (approved):
Great job on deduping behaviour here :)

PR #1152, opened 2022-07-26
Grouping access endpoint

sjaanus:
Added new endpoint that returns all users/groups with minimal fields.

### DIFF
--- a/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
+++ b/src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
@@ -110,7 +110,7 @@ export const ProjectAccessAssign = ({
     const { refetchProjectAccess } = useProjectAccess(projectId);
     const { addAccessToProject, changeUserRole, changeGroupRole, loading } =
         useProjectApi();
-    const { users = [], groups = [] } = useAccess();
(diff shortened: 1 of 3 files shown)


src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
```diff
@@ -109,8 +110,7 @@ export const ProjectAccessAssign = ({
     const { refetchProjectAccess } = useProjectAccess(projectId);
     const { addAccessToProject, changeUserRole, changeGroupRole, loading } =
         useProjectApi();
-    const { users = [] } = useUsers();
-    const { groups = [] } = useGroups();
+    const { users = [], groups = [] } = useAccess();
```
FredrikOseberg, line 113: Is this default needed? It seems to me that it's already setting a default in the useAccess hook.

FredrikOseberg (changes requested):

src/component/project/ProjectAccess/ProjectAccessAssign/ProjectAccessAssign.tsx
sjaanus: It is not, good catch.

FredrikOseberg (approved):

PR #1164, opened 2022-07-28
Fix: Refetch immutable toggle when adding strategy

Tymek:
- Fix a bug where it's not possible to edit a quickly added strategy, because of cached form data.
- Fix style of gradual rollout percentage indicator


nunogois (approved):
LGTM

PR #1190, opened 2022-08-02
Feat/groups refinement

sjaanus:
Refinements

1. If we have no groups lets add an empty state with a nice button 😃 (something similar we have on segments)
2. Verify that we don't have this toast message when there are no groups
3. Name should be required and highlighted if the group name exist (before pressing the create group button) → similar to the flow when we create a project
4. Add a hover state for the group card→ the same hover state that we use for project cards
5. We should align the avatars at the bottom of the card all the time
6. Lets adjust the avatars to 28px → to gain a bit more space
7. Improve the pop-over when hover a user
8. When we hover the project badge and it has project, add improvements
9. The options icon should stay at the top all the time even when we have a longer title 
10. update text from Remove group → Delete group
11. align the labels to the left
12. If there are no Project permissions → don't display the title
13. Same goes for Environment permissions → if none → don't display the title


src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
```diff
@@ -175,8 +175,16 @@ export const ProjectAccessTable: VFC = () => {
                         .sort()
                         .reverse()[0];
                 },
-                Cell: ({ value }: { value: Date }) => (
-                    <TimeAgoCell value={value} emptyText="Never logged" />
+                Cell: ({ value, row: { original: row } }: any) => (
+                    <ConditionallyRender
+                        condition={row.type === ENTITY_TYPE.USER}
```
nunogois, line 180: If we check the accessor above, in case it's a group, there's an attempt of fetching the most recent login from a user belonging to that group. However if this is a condition we want to add, we can also simplify the accessor.

src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
```diff
@@ -55,30 +71,32 @@ const StyledCounterDescription = styled('span')(({ theme }) => ({
     marginLeft: theme.spacing(1),
 }));
 
-const ProjectBadgeContainer = styled('div')(() => ({}));
+const ProjectBadgeContainer = styled('div')(() => ({
+    width: '50%',
```
nunogois, line 75: If we're going with this approach, I think we'll also need this:
```suggestion
    display: 'flex',
    justifyContent: 'end',
    width: '50%',
```
Otherwise the project badges get misaligned:
image

src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
```diff
@@ -20,9 +20,21 @@ const StyledGroupCard = styled('aside')(({ theme }) => ({
...
     [theme.breakpoints.up('md')]: {
         padding: theme.spacing(4),
     },
+    '&:hover': {
+        transition: 'background-color 0.2s ease-in-out',
+        backgroundColor: theme.palette.neutral.light,
+    },
+}));
+
+const StyledTitleRow = styled('div')(() => ({
```
nunogois, line 34: Maybe this could also extend `StyledRow` (like StyledBottomRow is doing) instead of `div`? Only thing that changes is `alignItems`.

src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
```diff
@@ -92,13 +110,27 @@ export const GroupCard = ({ group }: IGroupCardProps) => {
...
+                                        <Badge
+                                            onClick={e => {
+                                                e.preventDefault();
+                                                navigate(
+                                                    `/projects/${project}/access`
+                                                );
+                                            }}
+                                            color="secondary"
+                                            icon={<TopicOutlinedIcon />}
+                                            sx={{ marginRight: 0.5 }}
```
nunogois, line 129: Might be worth creating a `StyledBadge` instead of using `sx`: https://mui.com/system/basics/#performance-tradeoff

nunogois (changes requested):
Some really nice changes, but I think it's worth having a second look at some things before merging.

src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
sjaanus: I went with maxWidth and that should cover it?

src/component/admin/groups/CreateGroup/CreateGroup.tsx
```diff
@@ -43,8 +44,12 @@ export const CreateGroup = () => {
                 confetti: true,
                 type: 'success',
             });
-        } catch (error: unknown) {
-            setToastApiError(formatUnknownError(error));
+        } catch (error: any) {
+            if (error.body.name === 'NameExistsError') {
+                setErrors(prev => ({ ...prev, name: error.message }));
```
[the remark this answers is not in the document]
sjaanus: I think setErrors is only for field validations on the form. That is the only field that is validated.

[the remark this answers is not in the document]
sjaanus: I think only places that uniqueness is validated are using `/validate` endpoint in the middle, which I think is too much for this component.

src/component/admin/groups/GroupUserRoleCell/GroupUserRoleCell.tsx
```diff
@@ -3,21 +3,20 @@ import { ConditionallyRender } from 'component/common/ConditionallyRender/Condit
 import { TextCell } from 'component/common/Table/cells/TextCell/TextCell';
 import { Role } from 'interfaces/group';
 
-const StyledBadge = styled('div')(({ theme }) => ({
+export const StyledMemberBadge = styled('div')(({ theme }) => ({
```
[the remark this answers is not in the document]
sjaanus: Good catch. Done.

src/component/admin/groups/GroupsList/GroupCard/GroupCardAvatars/GroupPopover/GroupPopover.tsx
```diff
@@ -1,21 +1,19 @@
-import { Badge, Popover, styled } from '@mui/material';
+import { Popover, styled } from '@mui/material';
 import { IGroupUser, Role } from 'interfaces/group';
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
-import { Badge as StyledBadge } from 'component/common/Badge/Badge';
+import { Badge } from 'component/common/Badge/Badge';
 import StarIcon from '@mui/icons-material/Star';
+import { StyledMemberBadge } from '../../../../GroupUserRoleCell/GroupUserRoleCell';
```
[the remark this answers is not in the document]
sjaanus: Done.

src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
sjaanus: Good one. Changed.

src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
sjaanus: Done

src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
sjaanus: Talked with Nico, we will leave it as is, but how could we simplify?

src/component/admin/groups/CreateGroup/CreateGroup.tsx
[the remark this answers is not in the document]
nunogois: I see your point. 
Should we at least add `NameExistsError` as a const / enum somewhere that can be imported instead of using the literal string?
Or maybe create a method similar to `formatUnknownError` that returns both error type and message from an unknown error to clean this up a bit and be a bit more standardized.

src/component/admin/groups/GroupUserRoleCell/GroupUserRoleCell.tsx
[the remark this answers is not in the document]
nunogois: Nice, this component now looks super clean!

src/component/admin/groups/GroupUserRoleCell/GroupUserRoleCell.tsx
```diff
@@ -35,8 +21,12 @@ export const GroupUserRoleCell = ({
     const renderBadge = () => (
         <ConditionallyRender
             condition={value === Role.Member}
-            show={<StyledBadge>{capitalize(value)}</StyledBadge>}
-            elseShow={<StyledOwnerBadge>{capitalize(value)}</StyledOwnerBadge>}
+            show={<Badge>{capitalize(value)}</Badge>}
+            elseShow={
+                <Badge color="success" icon={<StyledPopupStar />}>
```
nunogois, line 26: We might not want `icon={<StyledPopupStar />}` in this case. 
Can we double check with @NicolaeUnleash just to be sure?

src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
nunogois: If we don't care about groups last login info, I guess we can do something like:
```
accessor: (row: IProjectAccess) => {
    if (row.type === ENTITY_TYPE.USER) {
        const userRow = row.entity as IUser;
        return userRow.seenAt || '';
    }
},
```
Removing the groups logic from the accessor, since we are not displaying it anyways.

nunogois (changes requested):
Last few conversations left to resolve.

src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
```diff
@@ -92,13 +112,26 @@ export const GroupCard = ({ group }: IGroupCardProps) => {
                             <ConditionallyRender
                                 condition={group.projects.length > 0}
                                 show={group.projects.map(project => (
-                                    <Badge
-                                        color="secondary"
-                                        icon={<TopicOutlinedIcon />}
-                                        sx={{ marginRight: 0.5 }}
+                                    <Tooltip
+                                        key={group.id}
```
nunogois, line 116: One group might have multiple projects, so I think we should use a project-bound identifier instead here.
```suggestion
                                        key={project}
```

src/component/project/ProjectAccess/ProjectAccessTable/ProjectAccessTable.tsx
sjaanus: RIght, now I understand. The logic was already working. I think lets keep it as it was then.

src/component/admin/groups/GroupsList/GroupCard/GroupCard.tsx
sjaanus: Ah seems to be typo. Nice catch.

src/component/admin/groups/CreateGroup/CreateGroup.tsx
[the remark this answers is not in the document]
sjaanus: Removed for now, because could not solve it prettily.

src/component/admin/groups/GroupUserRoleCell/GroupUserRoleCell.tsx
sjaanus: Yep we changed it everywhere

nunogois (approved):
LGTM

PR #1192, opened 2022-08-03
fix: validate feature strategy parameters 

olav:
This PR adds more validation to the feature strategy form. 

We now check that required strategy parameters have been filled in, on change and on submit.

To make this validation easier to work with, the PR also fills in more default values than before for the required parameters.

Part of https://linear.app/unleash/issue/1-99/custom-strategies-not-working-properly

Updates https://github.com/Unleash/unleash/issues/1879


Tymek:
@olav please remember to add description :)

FredrikOseberg:
> @olav please remember to add description :)

@olav Could you also add the issue this relates to on the public roadmap please?

src/component/feature/FeatureStrategy/FeatureStrategyCreate/FeatureStrategyCreate.tsx
```diff
@@ -46,8 +48,8 @@ export const FeatureStrategyCreate = () => {
     );
 
     useEffect(() => {
-        // Fill in the default values once the strategies have been fetched.
-        setStrategy(getStrategyObject(strategies, strategyName, featureId));
+        const strategy = strategies.find(s => s.name === strategyName);
```
FredrikOseberg, line 51: Please avoid one letter variables.

src/component/feature/FeatureStrategy/FeatureStrategyForm/FeatureStrategyForm.tsx
```diff
@@ -45,44 +53,78 @@ export const FeatureStrategyForm = ({
...
+    if (uiConfigError) {
+        throw uiConfigError;
+    }
+
+    if (uiConfigLoading || !strategyDefinition) {
+        return null;
+    }
+
+    const findParameterDefinition = (name: string): IStrategyParameter => {
+        return strategyDefinition.parameters.find(p => p.name === name)!;
```
FredrikOseberg, line 85:
```suggestion
        return strategyDefinition.parameters.find(parameter => parameter.name === name)!;
```

src/component/feature/FeatureStrategy/FeatureStrategyForm/FeatureStrategyForm.tsx
```diff
@@ -45,44 +53,78 @@ export const FeatureStrategyForm = ({
...
+            return false;
+        } else {
+            errors.remove(name);
+            return true;
+        }
+    };
+
+    const validateAllParameters = (): boolean => {
+        return strategyDefinition.parameters
+            .map(p => validateParameter(p.name, strategy.parameters?.[p.name]))
```
FredrikOseberg, line 107: Please update this to full variable name so that scanning the code become easier.

src/component/feature/FeatureStrategy/FeatureStrategyType/FeatureStrategyType.tsx
```diff
@@ -1,46 +1,44 @@
...
     const { context } = useUnleashContext();
 
-    const strategyDefinition = strategies.find(definition => {
-        return definition.name === strategy.name;
-    });
-
-    const updateParameter = (field: string, value: string) => {
+    const updateParameter = (name: string, value: string) => {
         setStrategy(
             produce(draft => {
```
FredrikOseberg, line 34: As a note, we should share this pattern. Produce can help us avoid bugs.

src/component/feature/StrategyTypes/GeneralStrategy/GeneralStrategy.tsx
```diff
@@ -1,190 +1,47 @@
...
-                                )}
-                            </div>
-                        );
-                    }
-                }
-            )}
-        </div>
+        <StyledContainer>
+            {strategyDefinition.parameters.map((definition, index) => (
+                <div key={index}>
```
FredrikOseberg, line 34: is the objectId function on definition a better key?

src/hooks/useFormErrors.ts
```diff
@@ -0,0 +1,62 @@
...
+        },
+        [setErrors]
+    );
+
+    const some = useCallback(
+        (): boolean => Object.values(errors).some(Boolean),
+        [errors]
+    );
+
+    return useMemo(
```
FredrikOseberg, line 53: What is the value of using useMemo here when all of the references are stabilized by useCallback? Seems to me we're only adding overhead with this.

FredrikOseberg (changes requested):

src/component/feature/StrategyTypes/GeneralStrategy/GeneralStrategy.tsx
olav: I think `index` should be enough since the order of the parameters won't change.

src/hooks/useFormErrors.ts
olav: The idea was that the return value from this hook (`errors`) should be a stable reference, so that it can be passed to optimized components directly, and used in `useEffect` dependency arrays (instead of passing each fn separately).

FredrikOseberg: But are we using it in any use effects? Or is this a premature optimization? 

Would it make sense to put the entire errors object in a useEffect? My hunch is that you would rarely all of these methods inside the useEffect? If you use `errors.get()` or `errors.some` those would still be stable references or am I missing something here?

olav: True, not needed yet. I can remove the last `useMemo` and instead expand the function names to guide their usage.

FredrikOseberg (approved):
LGTM

PR #1193, opened 2022-08-03
refactor: port event pages to TS

olav:
In preparation for https://linear.app/unleash/issue/1-120/add-searchfilter-for-event-log


andreas-unleash (approved):
LGTM!

PR #1207, opened 2022-08-05
New project page header

sjaanus:
1. Added description to header
2. Added projectId to header
3. Added possiblity to delete project (Created reusable component)

### DIFF
--- a/src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
+++ b/src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
@@ -1,15 +1,14 @@
-import { Dialogue } from '../../../common/Dialogue/Dialogue';
+import { Dialogue } from 'component/common/Dialogue/Dialogue';
 import React from 'react';
-import { formatUnknownError } from '../../../../utils/formatUnknownError';
-import useProjectApi from '../../../../hooks/api/actions/useProjectApi/useProjectApi';
-import useProjects from '../../../../hooks/api/getters/useProjects/useProjects';
-import useToast from '../../../../hooks/useToast';
+import { formatUnknownError } from 'utils/formatUnknownError';
+import useProjectApi from 'hooks/api/actions/useProjectApi/useProjectApi';
+import useProjects from 'hooks/api/getters/useProjects/useProjects';
+import useToast from 'hooks/useToast';
 
 interface IDeleteProjectDialogueProps {
     project: string;
     open: boolean;
-    onClose?: (e: React.SyntheticEvent, reason?: string) => void;
-    onFinish?: () => void;
+    onClose?: () => void;
     onSuccess?: () => void;
 }
 
@@ -17,35 +16,33 @@ export const DeleteProjectDialogue = ({
     open,
     onClose,
     project,
-    onFinish,
     onSuccess,
 }: IDeleteProjectDialogueProps) => {
     const { deleteProject } = useProjectApi();
     const { refetch: refetchProjectOverview } = useProjects();
     const { setToastData, setToastApiError } = useToast();
+
+    const onClick = async (e: React.SyntheticEvent) => {
+        e.preventDefault();
+        try {
+            await deleteProject(project);
+            refetchProjectOverview();
+            setToastData({
+                title: 'Deleted project',
+                type: 'success',
+                text: 'Successfully deleted project',
+            });
+            onSuccess?.();
+        } catch (ex: unknown) {
+            setToastApiError(formatUnknownError(ex));
+        }
+        onClose?.();
+    };
+
     return (
         <Dialogue
             open={open}
-            onClick={async (e: React.SyntheticEvent) => {
-                e.preventDefault();
-                try {
-                    await deleteProject(project);
-                    refetchProjectOverview();
-                    setToastData({
-                        title: 'Deleted project',
-                        type: 'success',
-                        text: 'Successfully deleted project',
-                    });
-                    if (onSuccess) {
-                        onSuccess();
-                    }
-                } catch (ex: unknown) {
-                    setToastApiError(formatUnknownError(ex));
-                }
-                if (onFinish) {
-                    onFinish();
-                }
-            }}
+            onClick={onClick}
             onClose={onClose}
             title="Really delete project"
         />
--- a/src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
+++ b/src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
@@ -1,0 +1,53 @@
+import { Dialogue } from '../../../common/Dialogue/Dialogue';
+import React from 'react';
+import { formatUnknownError } from '../../../../utils/formatUnknownError';
+import useProjectApi from '../../../../hooks/api/actions/useProjectApi/useProjectApi';
+import useProjects from '../../../../hooks/api/getters/useProjects/useProjects';
+import useToast from '../../../../hooks/useToast';
+
+interface IDeleteProjectDialogueProps {
+    project: string;
+    open: boolean;
+    onClose?: (e: React.SyntheticEvent, reason?: string) => void;
+    onFinish?: () => void;
+    onSuccess?: () => void;
+}
+
+export const DeleteProjectDialogue = ({
+    open,
+    onClose,
+    project,
+    onFinish,
+    onSuccess,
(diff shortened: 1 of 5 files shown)


src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
```diff
@@ -0,0 +1,53 @@
+import { Dialogue } from '../../../common/Dialogue/Dialogue';
+import React from 'react';
+import { formatUnknownError } from '../../../../utils/formatUnknownError';
+import useProjectApi from '../../../../hooks/api/actions/useProjectApi/useProjectApi';
+import useProjects from '../../../../hooks/api/getters/useProjects/useProjects';
+import useToast from '../../../../hooks/useToast';
```
nunogois, line 6: Should avoid relative paths.

src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
```diff
@@ -0,0 +1,53 @@
...
+    onFinish,
+    onSuccess,
+}: IDeleteProjectDialogueProps) => {
+    const { deleteProject } = useProjectApi();
+    const { refetch: refetchProjectOverview } = useProjects();
+    const { setToastData, setToastApiError } = useToast();
+    return (
+        <Dialogue
+            open={open}
+            onClick={async (e: React.SyntheticEvent) => {
```
nunogois, line 29: Might be worth extracting this into a function / handler so it's more readable.

src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
```diff
@@ -0,0 +1,53 @@
...
+                        text: 'Successfully deleted project',
+                    });
+                    if (onSuccess) {
+                        onSuccess();
+                    }
+                } catch (ex: unknown) {
+                    setToastApiError(formatUnknownError(ex));
+                }
+                if (onFinish) {
+                    onFinish();
```
nunogois, line 46: I guess it depends on who you ask, since it might be trading readability for simplicity, but I would prefer `onFinish?.()` here instead of the condition: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Same on `onSuccess()` above.

src/component/project/Project/Project.tsx
```diff
@@ -4,21 +4,25 @@ import useLoading from 'hooks/useLoading';
...
 import { ConditionallyRender } from 'component/common/ConditionallyRender/ConditionallyRender';
 import { useStyles } from './Project.styles';
-import { Tab, Tabs } from '@mui/material';
-import { Edit } from '@mui/icons-material';
+import { styled, Tab, Tabs } from '@mui/material';
+import { Delete, Edit } from '@mui/icons-material';
 import useToast from 'hooks/useToast';
 import useQueryParams from 'hooks/useQueryParams';
-import { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
```
nunogois, line 11: `React` is an unused import.

src/component/project/Project/Project.tsx
```diff
@@ -80,26 +86,84 @@ const Project = () => {
         /* eslint-disable-next-line */
     }, []);
 
+    const StyledDiv = styled('div')(({ theme }) => ({
```
nunogois, line 89: Unused `theme`, can just do `() =>` instead.

src/component/project/Project/Project.tsx
```diff
@@ -80,26 +86,84 @@ const Project = () => {
...
+    }));
+
+    const StyledTitle = styled('span')(({ theme }) => ({
+        fontSize: theme.fontSizes.smallBody,
+        fontWeight: 'normal',
+    }));
+    const StyledText = styled(StyledTitle)(({ theme }) => ({
+        color: theme.palette.grey[800],
+    }));
+
```
nunogois, line 107: Styled components should be moved to the top of the file, outside the component.

src/component/project/Project/Project.tsx
```diff
@@ -80,26 +86,84 @@ const Project = () => {
...
+                            <PermissionIconButton
+                                permission={DELETE_PROJECT}
+                                projectId={projectId}
+                                sx={{
+                                    visibility: isOss() ? 'hidden' : 'visible',
+                                }}
+                                onClick={() => {
+                                    setShowDelDialog(true);
+                                }}
+                                tooltipProps={{ title: 'Remove project' }}
```
nunogois, line 161: I noticed we are using `remove` in some places and `delete` in others. 
Do you know if there is a reasoning or guideline we can follow?

src/component/project/ProjectCard/ProjectCard.tsx
```diff
@@ -4,21 +4,17 @@ import MoreVertIcon from '@mui/icons-material/MoreVert';
 import { ReactComponent as ProjectIcon } from 'assets/icons/projectIcon.svg';
 import React, { useState, SyntheticEvent, useContext } from 'react';
```
nunogois, line 5: Unused `React` import.

src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
```diff
@@ -0,0 +1,53 @@
...
+    onFinish?: () => void;
+    onSuccess?: () => void;
+}
+
+export const DeleteProjectDialogue = ({
+    open,
+    onClose,
+    project,
+    onFinish,
+    onSuccess,
```
nunogois, line 21: Having:
 - `onClose`
 - `onFinish`
 - `onSuccess`
 
Seems a bit confusing at first sight. Might be worth having a second look to check whether we can simplify this.

nunogois (changes requested):
Looks good, but as usual some small nitpicks and comments that might be worth checking out before merging.

src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
nunogois: Could a solution for both this and https://github.com/Unleash/unleash-frontend/pull/1207#discussion_r938826791 be to extract this as a `onDelete` prop that is provided by the parent components?

At that point we might not need a dedicated `DeleteProjectDialogue` component and instead just use a plain `Dialogue` 🤔 Something to think about.

src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
sjaanus: Done

src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
sjaanus: Moved it as separate function.

I dont think we should duplicate this on parent layer. Then we will have two almost identical methods on parent layer. The whole point of creating reusable component was to move this logic to where it belongs.

src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
sjaanus: I like it also. Done.

src/component/project/Project/Project.tsx
sjaanus: Done

src/component/project/Project/Project.tsx
sjaanus: Done

src/component/project/Project/Project.tsx
sjaanus: Done

src/component/project/Project/Project.tsx
sjaanus: I think we used to have `remove` alot, but now slowly moving to `delete` on everywhere. Changed this to delete also.

src/component/project/ProjectCard/ProjectCard.tsx
sjaanus: Fixed

src/component/project/Project/DeleteProject/DeleteProjectDialogue.tsx
sjaanus: Removed onFinish

nunogois (approved):
LGTM

PR #1209, opened 2022-08-08
refactor: improve feature sidebar tooltip placement

olav:


src/component/feature/FeatureView/FeatureOverview/FeatureOverviewEnvSwitches/FeatureOverviewEnvSwitches.tsx
```diff
@@ -37,15 +70,14 @@ const FeatureOverviewEnvSwitches = () => {
...
-            >
-                <h3 className={styles.header} data-loading>
-                    Feature toggle status
-                </h3>
-            </Tooltip>
+        <StyledContainer>
+            <StyledHeader data-loading>
+                Feature toggle status
+                <HelpIcon
+                    tooltip="Environments can be switched off for a single toggle. Resulting in all calls towards the toggle to return false."
```
Tymek, line 77: Looks like 1 sentence, unless "...toggle. It will result in all calls towards the toggle returning `false`."

olav: Ah nice catch. How about something like this?

> When a feature is switched off in an environment, it will always return false. When switched on, it will return true or false depending on its strategies.

Tymek (approved):

